PR #166, opened 2016-08-17
Guest checkout should be enabled by default.

### LINKED ISSUE #165: PayPal Guest Payments option is disabled
PayPal Guest Payments are now incorrectly disabled by default. The conditional [here](https://github.com/woothemes/woocommerce-gateway-paypal-express-checkout/blob/11cafbab51101e10bf8491893c55019467f6d728/includes/class-wc-gateway-ppec-settings.php#L117) needs to check against `true` rather than `false`.  This will allow the filter to disable guest payments when needed. 

@gedex

gedex:
Fixes the filter that should be true by default to set the 'Sole' param.

Fixes #165.


gedex:
@dwainm please review.

dwainm:
@gedex LGTM

PR #168, opened 2016-08-20
Added IPN handler.

### LINKED ISSUE #131: Add IPN handler
There's an setting `pp_woo_ipnUrl` that's stored in option. This should be auto-determined like PayPal standard instead of exposed to merchant to decide and we could borrow the same response handler from PayPal standard handler.

Related ticket: 429146-zd-woothemes

gedex:
Fixes #131.


gedex:
@roykho please review

roykho:
LGTM!

PR #177, opened 2016-09-06
Set PayPal Guest Checkout to enabled by default

### LINKED ISSUE #176: PayPal Guest Payments option is disabled on checkout page
#165 enabled PayPal Guest checkout by default on the cart page, but the checkout page still exhibits the old behavior.

mdmoore:
Extends to the changes from #166 to the checkout page.  

Fixes #176 

@gedex


dwainm:
@mdmoore it's preferred that you use strict comparison `true === `

PR #186, opened 2016-10-05
Added rcorson to contributors

hannahswain:


hannahswain:
@dwainm would you please check this? thanks!

Issue #187, opened 2016-10-13
Shipping address is overwritten, but why?


kopiro:
https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-checkout-handler.php#L217

Here, the shipping address is overwritten by the Paypal response.
The main question is, why?

If we set the `ADDROVERRIDE` param to `true` (as is), we can discard the address returned by Paypal.

https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-checkout-handler.php#L530

dwainm:
Hi @kopiro 

Please see discussions on #190 . With how the PayPal Express checkout extension works with WooCommerce, we do not require users to be logged in so we have no guaranteed address when confirming the order. This is the most common case for this plugin.

If you would like to use shipping details in the users WordPress account you should use one of the other PayPal options which we suggest will be more appropriate for this.

dwainm:
Some options for you to consider:
https://woocommerce.com/products/woocommerce-gateway-paypal-powered-by-braintree/
https://woocommerce.com/products/paypal-advanced/

kopiro:
Understood.
But, there is a little "bug" that is very hard to find, and of course it was my case.

I had a "custom checkout field" called "company", but you overwrite it :)

https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-checkout-handler.php#L148

I deleted that row, and everything fine, but maybe you can call it in another way, maybe `paypal_company`, something like that.

dwainm:
@kopiro renaming the field would require a refactor in other places as well so not something we'd prefer to do now.

PR #190, opened 2016-10-17
Shipping fix

dwainm:
#189 to be merged first.

This adds shipping details to the package after the user is authenticated by PayPal.

Here we assume that logged in user already have shipping details on the site.

@roykho please test.


roykho:
@dwainm - I believe we should remove the check if user is logged in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/190/commits/9a3abfde20cdca1b0f9ece7296598d6bf68d7ed2#diff-08920fe9607f6499fddae3f0128f8b14R664 because we always want to use the address from PayPal and not from WC. PayPal Express's convenience is to bypass having to enter any addresses so it assumes what you have in your PayPal account is the address you want to use and it even passes it back to WC for use. If you fallback to the address that's on WC for calculating shipping, it would not coincide with the shipping address shown on the order receive page.

dwainm:
@roykho understood, but how does it affect this: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/187

roykho:
@dwainm - In the code, it is set to always override but yet the returned address shown on the order received page is of PayPal's account address. So there are some discrepancies there already.

To me, allow override should never be on reason being people who uses PPEC wants an "express" checkout experience meaning they are most likely not even logged in at the time they click on the PayPal checkout button. Which then means no address is even passed to PayPal at this time. So if we leave "allow override" feature in, it would mean it only should work for logged in customer who has their shipping address set already. This also means we would need to create an option for this in admin (enable/disable). If enabled, it works only for existing customers ( logged in ) and if disabled, then PayPal address is used. Thoughts?

dwainm:
@roykho I feel that we should go with the perception that this is for 'express checkout'. In the more common case, a user will not be logged in, even if they have an account with the store. Login details are only required on the checkout page which this extension totally bypasses. If people want to use the store shipping details they should use another PayPal option.

Thoughts?

roykho:
@dwainm - yes I would agree on that. It is even stated in the PayPal docs that they also recommend a "confirmed" address which is from PayPal.

dwainm:
Great, will also remove that override option as part of the fix.

dwainm:
Fixed locally, waiting for the OK

roykho:
@dwainm - I think we may have forgotten the use case of when the customer checkouts out on the checkout page and not the cart page. Then it should probably be override.

dwainm:
@roykho right, the shipping needs to be overridden if the user has gone to the checkout page. I have updated the commit and removed the `$params['ADDROVERRIDE'] = '1';`

Things work as expected for me:
* When not logged in and no shipping is given, PayPal stored details are sent back to our store.
* When not logged in and shipping details are provided on checkout. Those details are sent to PayPal and are also returned on checkout to process.

gedex:
@dwainm @roykho is this ready to merge?

roykho:
@gedex no, I believe @dwainm still needs to push up his changes.

dwainm:
@roykho @gedex shipping fix now in master.

PR #202, opened 2016-11-01
Added link to reset current credentials to allow reconnection.

gedex:
Also fixes toggle to credential fields if merchant prefers to manually
enter their API credentials.

Resolves #201.

PR #203, opened 2016-11-01
Rename settings loader to 'load' as the instance already implies.

gedex:
Also added magic setter and save.

Based off of `issues/201/credentials-related-overhaul`, so need #202 to be merged first and change the base to `master`.


gedex:
@dwainm please review.

Issue #9, opened 2016-02-09
[P2][4.1] Add billing agreement / subscriptions support


allendav:
per sections 11.9, 12.3 and 12.4 of the requirements doc

gedex:
Note from the meeting: Use reference transaction instead of billing feature from PayPal.

PR #211, opened 2016-11-07
Unify NVP-params getter for EC-related requests.

### LINKED ISSUE #207: Removes code for parallel payment details support (always use n=0)
This extension only uses a single payment detail for each order, so no need to specify bucket of payment details or parallel payments. In short, each request will only use one bucket -- `PAYMENTREQUEST_0_` and `L_PAYMENTREQUEST_0_`. This will simplify our codebase and unify parameters builder before each NVP request. Currently it's scattered and some are duplicated across classes.

/cc @mikejolley @mikaey In case you've feedback regarding this

### LINKED ISSUE #208: Add unified NVP params getters for EC-related calls
Currently it's resides in checkout's handler, cart handler, and in settings wrapper. Since this is related to API call it's more appropriate to have the getter in client's class.

## Methods to expect

Signature would be: `get_{EC_method_name}_params`. For example there's `WC_Gateway_PPEC_Client::set_express_checkout` that does `SetExpressCheckout` call, then its parameters getter would be `WC_Gateway_PPEC_Client::get_set_express_checkout`.

```
WC_Gateway_PPEC_Client::get_set_express_checkout_params
WC_Gateway_PPEC_Client::get_do_express_checkout_payment_params
```

gedex:
Before this, NVP-params getters were scattered across classes, making it
hard to track params injection when new feature is going to be added,
e.g., [reference transaction for subscriptions support](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/9).

The best place for the getters would be in client, just like each
EC-related call.

Resolves #207.
Resolves #208.


gedex:
@dwainm please review.

gedex:
Yep, was thinking about that before, but not sure what would be the best name for that as `get_set_.._params` is to retrieve params for `set_...` method.

PR #217, opened 2016-11-23
Handle customer registration in checkout page.

### LINKED ISSUE #172: When guest checkout disabled user is returned to checkout with an error prompt to fill in password field
IF you have the Guest checkout disabled in WooCommerce settings and the user logs into PayPal via the paypal express button on the Cart page, it will drop the user to the checkout page on return complaining that the password field was not set.

https://snag.gy/0jYkHy.jpg

gedex:
This fixes issue when guest checkout disabled user is returned to
checkout with an error prompt to fill in password field.

Fixes #172.


gedex:
@dwainm please review

PR #220, opened 2016-11-23
Fixed tax not updated according to PayPal shipping address.

### LINKED ISSUE #183: Sales tax not updating correctly.
WooCommerce PayPal Express Checkout Gateway - Version 1.1.2
WooCommerce - Version 2.6.4
- Default Customer Location `Shop base address`
- Set your store with a tax local to your base location (eg. Ohio state tax)
- Calculate Tax Based On `Customer Shipping Address`

Add items to your cart, then click the yellow "Checkout with PayPal" button. In the PayPal checkout window select a shipping address that does not meet the local tax you previously set up (eg. Florida).

**Expected**
The cart's sales tax is updated as no sales tax should be charge for Florida

**Actual**
The cart's sales tax does not update and the Ohio state tax is still shown.

It's important to note that this only happens if the customer's address does not fall under one of the sales taxes that you've set up. If you use an address that falls under a different tax, the cart will update to the proper tax.

### LINKED ISSUE #219: Minimize call to get_checkout_details
In checkout page there's multiple calls of `get_checkout_details`. Especially in [`paypal_billing_details`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-checkout-handler.php#L93-L115) and [`paypal_shipping_details`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-checkout-handler.php#L159-L171). This need to be cached in session / transient / object's property as `get_checkout_details` involves remote request.

gedex:
This change will update customer's shipping address in session based on selected shipping address in PayPal. Also resolves multiple calls to `get_checkout_details` from checkout page for billing and shipping sections.

Fixes #183.
Resolves #219.

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -97,7 +97,7 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			return;
 		}
 
-		// Sets customer shipping address in session based on checkout details =
+		// Sets customer shipping address in session based on checkout details
 		// from PayPal.
 		WC()->customer->set_shipping_country( $checkout_details->payments[0]->shipping_address->getCountry() );
 		WC()->customer->set_shipping_state( $checkout_details->payments[0]->shipping_address->getState() );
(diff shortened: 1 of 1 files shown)


gedex:
@roykho please review

PR #223, opened 2016-11-28
do not use the shipping address if it is set to false.

dwainm:
@gedex please review.


gedex:
Is there any issue pop up for this or you encountered this locally?

dwainm:
@gedex I found this while testing a bookings partial payment locally.

gedex:
@dwainm cool

PR #253, opened 2017-04-11
Disable gateway when the cart total is zero

### LINKED ISSUE #250: We need to hide PPEC when cart amount is zero
So while testing account funds, the total cart amount became zero as it is paid for by the funds in my account. In situations like this or if product is free, the PPEC button should be hidden.

Otherwise you get this error which is also not very user friendly.

allendav:
Fixes #250 

To Test:
* Enable and set up the gateway, including enabling it on checkout
* Add an item to the cart
* Ensure the gateway presents on both the cart and checkout pages
* Add a coupon to the cart that zeros out the cart
* Ensure the gateway is NOT presented on the cart nor checkout pages

### DIFF
--- a/woocommerce-gateway-paypal-express-checkout.php
+++ b/woocommerce-gateway-paypal-express-checkout.php
@@ -25,6 +25,10 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly
 }
 
+if ( ! defined( 'WC_GATEWAY_PPEC_VERSION' ) ) {
+	define( 'WC_GATEWAY_PPEC_VERSION', '1.2.1' );
+}
+
 /**
  * Return instance of WC_Gateway_PPEC_Plugin.
  *
@@ -36,7 +40,7 @@ function wc_gateway_ppec() {
 	if ( ! isset( $plugin ) ) {
 		require_once( 'includes/class-wc-gateway-ppec-plugin.php' );
 
-		$plugin = new WC_Gateway_PPEC_Plugin( __FILE__, '1.2.1' );
+		$plugin = new WC_Gateway_PPEC_Plugin( __FILE__, WC_GATEWAY_PPEC_VERSION );
 	}
 
 	return $plugin;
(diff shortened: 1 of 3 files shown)


woocommerce-gateway-paypal-express-checkout.php
```diff
@@ -36,7 +36,7 @@ function wc_gateway_ppec() {
 	if ( ! isset( $plugin ) ) {
 		require_once( 'includes/class-wc-gateway-ppec-plugin.php' );
 
-		$plugin = new WC_Gateway_PPEC_Plugin( __FILE__, '1.2.0' );
+		$plugin = new WC_Gateway_PPEC_Plugin( __FILE__, '1.2.1' );
```
roykho: Perhaps we should define a global version as per our standardization so that we don't need to change versioning in multiple places?

allendav: Agreed! https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/253/commits/d0f67b685a27f616d4da8cb6e89c44efbdddc574

woocommerce-gateway-paypal-express-checkout.php
```diff
@@ -25,6 +25,10 @@
 	exit; // Exit if accessed directly
 }
 
+if ( ! defined( 'WC_GATEWAY_PPEC_VERSION' ) ) {
+	define( 'WC_GATEWAY_PPEC_VERSION', '1.2.1' );
```
roykho: So for this if you recall, we agreed to just define it without first checking as we want it to break as soon as possible.

allendav: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/253/commits/845f9478e64726aadf3b479752b5e193cf344c0c How bow dah?

roykho: G2G! :)

PR #267, opened 2017-05-28
use shipping discount instead of tax when adjustment negative

Gator92:
The tax amount cannot be negative, use shipping discount in this case for total mismatch workarounds.


bor0:
@Gator92 Thanks for your contribution! Could you please provide a test case where this is fix applicable?

Gator92:
@bor0 I came across this when working on a shop where they implemented one of their discounts using the fee API. This would be applicable when there is any type of negative adjustment that isn't reflected in get_discounts.

PR #272, opened 2017-06-01
Do not process orders automatically

### LINKED ISSUE #260: Payment automatically goes to complete when payment action set to Authorize
558730-zd-woothemes

#### What I expected

Payment should be on-hold and set to processing manually

#### What happened instead

Automatically goes through processing/complete

#### Steps to reproduce the issue

Payment action set to Authorize

Link to screenshot: https://cldup.com/yvn2xCDHA2.png

Make an order

Status automatically captures payment:

Link to screenshot: https://cldup.com/lRBHytBKUU.png 

zd-ticket-577631

bor0:
Fixes #260.

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec-paypal-request-handler.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec-paypal-request-handler.php
@@ -73,14 +73,7 @@ abstract class WC_Gateway_PPEC_PayPal_Request_Handler {
 	 */
 	protected function payment_complete( $order, $txn_id = '', $note = '' ) {
 		$order->add_order_note( $note );
-
-		if ( ! empty( $txn_id ) ) {
-			$order->set_transaction_id( $txn_id );
-		}
-
-		if ( ! $order->get_date_paid( 'edit' ) ) {
-			$order->set_date_paid( current_time( 'timestamp', true ) );
-		}
+		$order->payment_complete( $txn_id );
 	}
 
 	/**
--- a/includes/abstracts/abstract-wc-gateway-ppec-paypal-request-handler.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec-paypal-request-handler.php
@@ -73,7 +73,14 @@ abstract class WC_Gateway_PPEC_PayPal_Request_Handler {
 	 */
 	protected function payment_complete( $order, $txn_id = '', $note = '' ) {
 		$order->add_order_note( $note );
-		$order->payment_complete( $txn_id );
+
+		if ( ! empty( $txn_id ) ) {
+			$order->set_transaction_id( $txn_id );
+		}
+
+		if ( ! $order->get_date_paid( 'edit' ) ) {
+			$order->set_date_paid( current_time( 'timestamp', true ) );
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


gedex:
AFAIK, if payment action is set to Authorize, you won't get pinged via PP IPN. The order status is updated by IPN handler, so I think it should remains in on-hold.

gedex:
Ok, i just tested this with master and apparently it updates the status :/. Testing your branch.

gedex:
@bor0 I think the culprit of this is [sandbox check on IPN handler](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-ipn-handler.php#L100-L102). PayPal standard in core has the same [check](https://github.com/woocommerce/woocommerce/blob/master/includes/gateways/paypal/includes/class-wc-gateway-paypal-ipn-handler.php#L57-L60). This won't affect live transaction. I'm not sure the reasoning behind that. @mikejolley did you remember why we need to set payment status to completed (in IPN handler) when sandbox is enabled?

The change introduced in this PR will affect transaction when payment action is set to **Sale** (authorize and capture). So, i suggest we either remove the sandbox check or add another check based on payment action setting.

@yukikatayama @mdmoore does the issue happens on live mode? I imagine if that's the case it will generate lots of tickets.

mikejolley:
@gedex A longstanding issue in PayPal IPN was that it was not possible to get 'completed' status back - only pending. So to work around this and test payments for real we added the test_ipn check to workaround it.

It only affects sandbox and should not affect live. It would be better to not work around it in other places and just accept that in testing with Sandbox, all transactions are completed.

bor0:
Thanks for confirming @mikejolley!

Waiting for @yukikatayama / @mdmoore to confirm that it only happens on sandbox before closing it as wont fix.

mikejolley:
@gedex @bor0 Or we could change the core code to only flip that status if not using authorisations?

bor0:
It would have to be both core change (for PayPal) and change in this extension (PPEC), right? Or did you mean changing some other part?

mikejolley:
@bor0 Just the core test_ipn check

bor0:
@mikejolley but we have that check both in
[Extension](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-ipn-handler.php#L100)
and
[Core](https://github.com/woocommerce/woocommerce/blob/master/includes/gateways/paypal/includes/class-wc-gateway-paypal-ipn-handler.php#L58)

So just confirming it needs both changes on the extension and core?

mikejolley:
Ah yes. I didn't realise PPEC had IPN too.

yukikatayama:
Update from @mdmoore and my testings - in Live mode, the correct behavior of going to on-hold is working! :) However when changing status to processing/complete, transaction is not made. Details [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/260#issuecomment-305541804).

bor0:
@gedex / @mikejolley another round of review please? (Along with https://github.com/woocommerce/woocommerce/pull/15402)

Issue #194, opened 2016-10-25
[P2][4.4] Add EC button on the product page


gedex:
This enables the merchant to do a one-click checkout from the product page. This is available in the other competitive plugins as well.
## Details
- On gateway setting, provides option (checkbox) to enable this
- On single product page, render the EC button below add to cart secion

bor0:
@mattyza @gedex I am wondering how should the flow be on this one. Should the PayPal popup appear on the product page or should we redirect them to the cart page and then show the popup?

PR #276, opened 2017-06-06
Implement EC button on product page

bor0:
Implements #194.

### DIFF
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -1,5 +1,5 @@
 /* global wc_ppec_context */
-;(function ( $, window, document ) {
+;(function( $, window, document ) {
 	'use strict';
 
 	var get_attributes = function() {
@@ -13,12 +13,12 @@
 			var value	  = $( this ).val() || '';
 
 			if ( value.length > 0 ) {
-				chosen ++;
+				chosen++;
 			}
 
-			count ++;
+			count++;
 			data[ attribute_name ] = value;
-		});
+		} );
 
 		return {
 			'count'      : count,
@@ -27,7 +27,7 @@
 		};
 	};
 
-	$( '#woo_pp_ec_button' ).click( function ( event ) {
+	$( '#woo_pp_ec_button' ).click( function( event ) {
 		event.preventDefault();
 
 		var data = {
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -1,0 +1,51 @@
+/* global wc_ppec_context */
+;(function ( $, window, document ) {
+	'use strict';
+
+	var get_attributes = function() {
+		var select = $( '.variations_form' ).find( '.variations select' ),
+			data   = {},
+			count  = 0,
+			chosen = 0;
+
+		select.each( function() {
+			var attribute_name = $( this ).data( 'attribute_name' ) || $( this ).attr( 'name' );
+			var value	  = $( this ).val() || '';
+
+			if ( value.length > 0 ) {
+				chosen ++;
+			}
+
+			count ++;
+			data[ attribute_name ] = value;
+		});
+
+		return {
+			'count'      : count,
+			'chosenCount': chosen,
+			'data'       : data
+		};
+	};
+
+	$( '#woo_pp_ec_button' ).click( function ( event ) {
+		event.preventDefault();
+
+		var data = {
+			'nonce':      wc_ppec_context.generate_cart_nonce,
+			'qty':        $( '.quantity .qty' ).val(),
+			'attributes': $( '.variations_form' ).length ? get_attributes().data : []
+		};
+
+		var href = $(this).attr( 'href' );
+
+		$.ajax({
+			type:    'POST',
+			data:    data,
+			url:     wc_ppec_context.ajaxurl,
+			success: function( response ) {
+				window.location.href = href;
+			}
+		});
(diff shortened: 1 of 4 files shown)


assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -0,0 +1,51 @@
...
+			data   = {},
+			count  = 0,
+			chosen = 0;
+
+		select.each( function() {
+			var attribute_name = $( this ).data( 'attribute_name' ) || $( this ).attr( 'name' );
+			var value	  = $( this ).val() || '';
+
+			if ( value.length > 0 ) {
+				chosen ++;
```
gedex: why there's a space before increment operator?

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -0,0 +1,51 @@
...
+		});
+
+		return {
+			'count'      : count,
+			'chosenCount': chosen,
+			'data'       : data
+		};
+	};
+
+	$( '#woo_pp_ec_button' ).click( function ( event ) {
```
gedex: space between `function` and `(`?

includes/settings/settings-ppec.php
```diff
@@ -376,4 +376,11 @@
 			'drop' => __( 'Do not send line items to PayPal', 'woocommerce-gateway-paypal-express-checkout' ),
 		),
 	),
+	'checkout_product' => array(
```
gedex: can we move this under **Display Settings** section instead of **Advanced Settings** ?

includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -38,6 +43,83 @@ public function before_cart_totals() {
...
+		/**
+		 * If this page is single product page, we need to simulate
+		 * adding the product to the cart taken account if it is a
+		 * simple or variable product.
+		 */
+		if ( is_single() ) {
+			// First empty the cart to prevent wrong calculation.
+			WC()->cart->empty_cart();
+
+			if ( 'variable' === $product->get_type() && isset( $_POST['attributes'] ) ) {
```
gedex: not a biggie, there's `$product->is_type( 'variable' )` if you want to use that.

includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -38,6 +43,83 @@ public function before_cart_totals() {
...
+
+		$product = wc_get_product( $post->ID );
+		$qty     = ! isset( $_POST['qty'] ) ? 1 : absint( $_POST['qty'] );
+
+		/**
+		 * If this page is single product page, we need to simulate
+		 * adding the product to the cart taken account if it is a
+		 * simple or variable product.
+		 */
+		if ( is_single() ) {
```
gedex: why don't use `is_product()` ?

includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -38,6 +43,83 @@ public function before_cart_totals() {
...
+			// First empty the cart to prevent wrong calculation.
+			WC()->cart->empty_cart();
```
gedex: can we make it works without emptying the cart? E.g. a customer has put A and B in the cart then goes to C and then clicking the button from there?

includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -38,6 +43,83 @@ public function before_cart_totals() {
...
+
+	/**
+	 * Display paypal button on the product page.
+	 *
+	 * @since 1.4.0
+	 */
+	public function display_paypal_button_product() {
+		$gateways = WC()->payment_gateways->get_available_payment_gateways();
+
+		if ( ! is_single() || ! isset( $gateways['ppec_paypal'] ) ) {
```
gedex: not a biggie and we can skip this for. This won't work on a page with `[product_page]` shortcode.

gedex:
@bor0 When I tested this two things that I noticed:

* It reset the cart. There's a line where a mentioned about this too.
* When button is clicked. it uses redirect experience rather than in-context (dialog that popup like when you clicked the button from cart page). If there's no way to make in-context experience works from this single product, I'm fine with that.

bor0:
@gedex addressed comments, one more round please. Thanks!

gedex:
Some test results:

* Multiple items in the cart works.
* In-context experience (popup) doesn't work when clicking from single product page. For me, it's not a blocker.

includes/settings/settings-ppec.php
```diff
@@ -324,6 +324,13 @@
 		'desc_tip'    => true,
 		'description' => __( 'This enables PayPal Credit, which displays a PayPal Credit button next to the Express Checkout button. PayPal Express Checkout lets you give customers access to financing through PayPal Credit® - at no additional cost to you. You get paid up front, even though customers have more time to pay. A pre-integrated payment button shows up next to the PayPal Button, and lets customers pay quickly with PayPal Credit®.', 'woocommerce-gateway-paypal-express-checkout' ),
 	),
+	'checkout_product' => array(
```
gedex: can we make a better setting's name here? I don't mind with long name like `checkout_on_single_product_enabled` or something like that.

bor0:
@gedex, updated setting name.

Regarding
> In-context experience (popup) doesn't work when clicking from single product page. For me, it's not a blocker.
I think it's fine because this is the same experience if you hover on the cart and checkout from there. It's re-using the same code instead of duplicating it. We can rework in the future if required.

PR #277, opened 2017-06-07
Always send shipping requirement to PayPal

### LINKED ISSUE #249: If there are no shipping options in WooCommerce, PayPal doesn't pass a shipping address to WC
A customer found that if they didn't enable any shipping options, PayPal wouldn't pass a shipping address back to the Store: https://wordpress.org/support/topic/no-billing-adress-back/page/2/

They needed to set shipping as free, rather than omit shipping altogether, but it's worth considering at some point.

bor0:
Fixes #249.


gedex:
@bor0 please resolve the conflict

bor0:
@gedex 👍

PR #278, opened 2017-06-07
Option to show/hide checkout on cart page

### LINKED ISSUE #173: Feature: Option to disable checkout with PayPal button on Cart page
On some sites the cart page is finely tuned to funnel users through to the checkout process via the existing Proceed To Checkout button.

In these cases it might be desirable to disable the checkout with PayPal button on the Cart page altogether since you can still use PayPal express quite effectively from the Checkout page.

bor0:
Fixes #173.

### DIFF
--- a/includes/settings/settings-ppec.php
+++ b/includes/settings/settings-ppec.php
@@ -263,6 +263,14 @@ return array(
 			'large'  => __( 'Large', 'woocommerce-gateway-paypal-express-checkout' ),
 		),
 	),
+	'cart_checkout' => array(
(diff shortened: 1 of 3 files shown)


includes/settings/settings-ppec.php
```diff
@@ -263,6 +263,14 @@
 			'large'  => __( 'Large', 'woocommerce-gateway-paypal-express-checkout' ),
 		),
 	),
+	'cart_checkout' => array(
```
gedex: can we make the name derives the boolean state from its type and possible values. Something like `cart_checkout_enabled`.

bor0:
@gedex updated name, thanks!

PR #281, opened 2017-06-09
Do not append version string to checkout.js

### LINKED ISSUE #280: Express Checkout goes to fallback mode, redirects to Paypal site.
Starting Wednesday we noticed that the checkout process was redirecting to Paypal instead of handling it in the popup window. It is doing a fallback workflow because of a JS error. We contacted PayPal support and they informed us that the problem was due to the url for checkout.js having the version string attached. 

We modified the plugin to remove the version string and it is working fine now.

singlow:
Removing the version string resolves an error that forces a fallback workflow and redirect to paypal.com.

Resolves #280


bor0:
Thank you for your contribution, @singlow!

PR #287, opened 2017-06-28
Fix noshipping by still checking shipping on products

### LINKED ISSUE #286: NOSHIPPING=0 always.
After update to v1.4.0 we got an error from Paypal about wrong shipping address. The product is virtual and doesn't require shipping. I believe that's wrong behavior and needs to be  fixed.

bor0:
Fixes #286.

### DIFF
--- a/includes/class-wc-gateway-ppec-plugin.php
+++ b/includes/class-wc-gateway-ppec-plugin.php
@@ -372,4 +372,21 @@ class WC_Gateway_PPEC_Plugin {
 
 		return array_merge( $plugin_links, $links );
 	}
+
+	/**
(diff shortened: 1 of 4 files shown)


includes/class-wc-gateway-ppec-plugin.php
```diff
@@ -372,4 +372,21 @@ public function plugin_action_links( $links ) {
 
 		return array_merge( $plugin_links, $links );
 	}
+
+	/**
+	 * Check if shipping is needed for PayPal.
+	 *
+	 * @since 1.4.1
+	 *
```
roykho: Would be good to put a `@version` here per our standards doc.

PR #291, opened 2017-06-29
Change status to completed after capturing

### LINKED ISSUE #284: _paypal_status on Authorize transactions not updating to completed after capture
_paypal_status is always "pending".

Is this suppose to update to "completed" after the transaction is captured?

bor0:
Fixes #284.


creativeround:
Tested, but changed the order status to "Processing" not "Completed".

creativeround:
Not working, tested.  The order status is changing to "Processing" not "Completed".

Attached are the order notes.
order-notes

creativeround:
This is not a virtual product.

creativeround:
I did notice that when you change the order status to "Cancelled" on an order that is NOT captured, it will create an order note and set the status to "Failed".  Is there a way we can set this to "Cancelled".  "Failed" is not really a desired status, since the order status selected was "Cancelled".

Order Note:
"Payment voided via IPN. Order status changed from Cancelled to Failed."

PR #302, opened 2017-07-17
Abstract get_details for both cart and order to ensure proper calcula…

### LINKED ISSUE #268: 10413 error
Hi,

10413 error when executing paypal express in-page checkout after applying coupon code on cart page.

We are on The this version:
WooCommerce 3.0.1
PayPal for WooCommerce 1.4.4

Thank you.

bor0:
…tion

Fixes #268.

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -458,7 +458,7 @@ class WC_Gateway_PPEC_Client {
 			'items'             => $this->_get_paypal_line_items_from_cart(),
 		);
 
-		return $this->_get_details( $details, $discounts, $rounded_total, WC()->cart->total );
+		return $this->get_details( $details, $discounts, $rounded_total, WC()->cart->total );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/class-wc-gateway-ppec-client.php
```diff
@@ -458,72 +458,7 @@ protected function _get_details_from_cart() {
 			'items'             => $this->_get_paypal_line_items_from_cart(),
 		);
```
gedex: since you pulled out the logic of details builder into its own method, maybe we could bring this `$details` initialization too?

gedex: nvm. Currently, it's confusing already because we have `_get_details_from_cart` and `_get_details_from_order`.

roykho: So the only two things I can see is per our standardization spec we no longer prefix private/protected methods with `_` underscore. And the other thing perhaps is if we need to cast anything to `float` in the calculations in case PHP7 barks. Just a mention as it may or may not be necessary in your case.

PR #306, opened 2017-07-20
Remove OpenSSL warning when using LibreSSL

Caomhin:
LibreSSL forked from OpenSSL 1.0.1g and is therefore > OpenSSL 1.0.1 so
the warning is redundant.

Add check for LibreSSL in OPENSSL_VERSION_TEXT and make non-caputuring
to avoid breaking the version number check


bor0:
Thanks for your contribution!

PR #307, opened 2017-07-24
Update order meta with new transaction id

### LINKED ISSUE #304: Refunds not working on authorize then captured transactions
When you capture the charge, PayPal does the capture in a new transaction ID.
The plugin is not updating the previous transaction ID to the new "captured" transaction ID.

When you go to do a refund, it's trying to refund on authorize transaction ID, not the final charge/capture transaction ID.

I manually changed the post_meta "_transaction_id" to the final charge/capture transaction ID and the refund processed.

bor0:
Fixes #304.

### DIFF
--- a/includes/class-wc-gateway-ppec-admin-handler.php
+++ b/includes/class-wc-gateway-ppec-admin-handler.php
@@ -168,6 +168,11 @@ class WC_Gateway_PPEC_Admin_Handler {
 					$order->add_order_note( __( 'Unable to capture charge!', 'woocommerce-gateway-paypal-express-checkout' ) . ' ' . $result->get_error_message() );
 				} else {
 					update_post_meta( $order_id, '_paypal_status', ! empty( $trans_details['PAYMENTSTATUS'] ) ? $trans_details['PAYMENTSTATUS'] : 'completed' );
+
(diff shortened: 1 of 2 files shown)


includes/class-wc-gateway-ppec-admin-handler.php
```diff
@@ -168,6 +168,11 @@ public function capture_payment( $order_id ) {
 					$order->add_order_note( __( 'Unable to capture charge!', 'woocommerce-gateway-paypal-express-checkout' ) . ' ' . $result->get_error_message() );
 				} else {
 					update_post_meta( $order_id, '_paypal_status', ! empty( $trans_details['PAYMENTSTATUS'] ) ? $trans_details['PAYMENTSTATUS'] : 'completed' );
+
+					if ( ! empty( $trans_details['TRANSACTIONID'] ) ) {
+						update_post_meta( $order_id, '_paypal_status', $trans_details['TRANSACTIONID'] );
```
roykho: How come transactionid is going into "_paypal_status" meta? Shouldn't it be "_transaction_id" ?

bor0: Ha! Nice catch, copy & paste error

PR #310, opened 2017-08-12
Fix error by properly checking zero for floats

### LINKED ISSUE #309: 10431 Error – Item Amount Invalid at Checkout
I came across a bug when using the live environment, **and** ~~which wasn't manifesting itself using~~ the sandbox... I am seeing a [10431 error](https://developer.paypal.com/docs/classic/api/errors/#10431) on the checkout page.
It is also possibly related to using a coupon code since in my limited testing it occurred after adding a coupon.
I also don't get the error if I click the 'Check out with PayPal' button on the cart page, coupon code or not.

**Environment**
- WordPress `V4.8`
- WooCommerce `V3.1.1`
- WooCommerce PayPal Express Checkout Gateway `V1.4.2`

**Steps to Reproduce**
- Add an item worth $24.95 to the cart
- Apply a coupon code which gives a $4.95 discount
- Click the 'Proceed to checkout' button 
- Select PayPal Express payment option
- Click the 'Continue to payment' button
- If I remove the coupon, it works as expected

It is also reported over on the [WordPress Plugin page](https://wordpress.org/support/topic/paypal-error-10431/).

**Screenshot**
screen shot 2017-08-01 at 9 54 05 am

bor0:
Fixes #309.

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -547,7 +547,7 @@ class WC_Gateway_PPEC_Client {
 				// Add line item to make up different between WooCommerce
 				// calculations and PayPal calculations.
 				$diff = round( $details['total_item_amount'] - $rounded_total, $decimals );
-				if ( $diff != 0 ) {
+				if ( abs( $diff ) < 0.0001 ) {
 					$extra_line_item = $this->_get_extra_offset_line_item( $diff );
 
 					$details['items'][]            = $extra_line_item;
@@ -603,7 +603,7 @@ class WC_Gateway_PPEC_Client {
 			$lisum += $values['quantity'] * $values['amount'];
 		}
 
-		if ( 0 !== $lisum ) {
+		if ( abs( $lisum ) < 0.0001 ) {
 			$details['items'][] = $this->_get_extra_offset_line_item( $details['total_item_amount'] - $lisum );
 		}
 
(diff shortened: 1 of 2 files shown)


hamoderr:
I know I was not invited to review. My acceptance is just an indication that that making these changes locally has resolved the issue for me.

bor0:
Your local testings imply your invitation for review. Glad it's resolved your issue.

Thanks!

includes/class-wc-gateway-ppec-client.php
```diff
@@ -547,7 +547,7 @@ protected function get_details( $details, $discounts, $rounded_total, $total ) {
 				// Add line item to make up different between WooCommerce
 				// calculations and PayPal calculations.
 				$diff = round( $details['total_item_amount'] - $rounded_total, $decimals );
-				if ( $diff != 0 ) {
+				if ( abs( $diff ) < 0.0001 ) {
```
gedex: what's the reasoning to choose the number `0.0001` ? Is it mentioned in PPEC doc?

bor0: It uses the formula `|a - b| < e` to see how close a and b are for a small e (the smaller the better). In this case b = 0 and e = 0.0001.

As far as I checked debug logs we were sending PP a very small number which was caused by the calculation above on floats, so PP didn't like that number.

PR #315, opened 2017-09-08
Undefined item issue resolve #314

### LINKED ISSUE #314: Notice: Undefined index: items class-wc-gateway-ppec-client.php on line 602
Hi

If I set Subtotal Mismatch Behaviour as *Do not send line items to PayPal* when the user lands back on checkout complete page I am getting these warnings.

```
Notice: Undefined index: items in wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-client.php on line 602

Warning: Invalid argument supplied for foreach() in wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-client.php on line 602
```
I am not entirely familiar with the logic of not returning items and if that messes anything up down the line. So don't want to mess making a pull request. But it seems quite easy to dismiss this warning with a `isset($details['items'])`?

Hope we can fix soon.

mitraval192:
Related to issue #314

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -599,8 +599,10 @@ class WC_Gateway_PPEC_Client {
 
 		$lisum = 0;
 
-		foreach ( $details['items'] as $li => $values ) {
-			$lisum += $values['quantity'] * $values['amount'];
+		if ( ! empty( $details['items'] ) ) {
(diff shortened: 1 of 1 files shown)


bor0:
Thank you for your contribution @mitraval192!

mitraval192:
@bor0 Thanks for the Merged.
Please suggest me if we need to add anything to it.

PR #331, opened 2017-11-13
Normalize states, add support for Ireland

### LINKED ISSUE #323: Irish County names don't match
When I checkout with PayPal Express using an Irish address, I'm required to select a county. However PayPal adds the `CO` prefix in front of the name.

So when I do the purchase I get the following error:

```
Billing County is not valid. Please enter one of the following: Clare, Cork, Cavan, Carlow, Donegal, Dublin, Galway, Kildare, Kilkenny, Kerry, Longford, Louth, Limerick, Leitrim, Laois, Meath, Monaghan, Mayo, Offaly, Roscommon, Sligo, Tipperary, Waterford, Westmeath, Wicklow, Wexford
```

This is the address as it is returned from PayPal Express:

```
    [SHIPTONAME] => Test Ireland
    [SHIPTOSTREET] => Address
    [SHIPTOCITY] => City
    [SHIPTOSTATE] => CO GALWAY
    [SHIPTOCOUNTRYCODE] => IE
    [SHIPTOCOUNTRYNAME] => Ireland
```

The county in WooCommerce is defined without the prefix. See: https://github.com/woocommerce/woocommerce/blob/3.2.1/i18n/states/IE.php#L23

image

This blocks all payments which ship to an Irish address.

Affected ticket: 740831-zd-woothemes

bor0:
Fixes #323.


bor0:
@mikkamp note I haven't been able to test this, my sandbox has some issues around PP creds. Please let me know how it goes. Thanks!

gedex:
Tested and works, but got notice which seems have been there before this change:

```
[13-Nov-2017 14:51:52 UTC] PHP Notice:  Undefined property: PayPal_Address::$_countryCode in /srv/www/wordpress-default/wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-address.php on line 360
```

it should be `$this->_country`.

mikkamp:
@bor0 Wasn't able to test this until now.

The changes do work and allow me to complete the checkout when I select a county. Although I noticed a few more problems.

After making the purchase a second address is saved in PayPal using the WooCommerce state code instead of the CO Galway.

image

This isn't a real problem, but just a bit confusing since the county code is not familiar to most users.

Also the PayPal option isn't just the county but they use "County / Dublin postal district". So it can contain all the following options

```
CO CARLOW
CO CAVAN
CO CLARE
CO CORK
CO DONEGAL
CO DUBLIN
CO GALWAY
CO KERRY
CO KILDARE
CO KILKENNY
CO LAOIS
CO LEITRIM
CO LIMERICK
CO LONGFORD
CO LOUTH
CO MAYO
CO MEATH
CO MONAGHAN
CO OFFALY
CO ROSCOMMON
CO SLIGO
CO TIPPERARY
CO WATERFORD
CO WESTMEATH
CO WEXFORD
CO WICKLOW
DUBLIN 1
DUBLIN 2
DUBLIN 3
DUBLIN 4
DUBLIN 5
DUBLIN 6
DUBLIN 6W
DUBLIN 7
DUBLIN 8
DUBLIN 9
DUBLIN 10
DUBLIN 11
DUBLIN 12
DUBLIN 13
DUBLIN 14
DUBLIN 15
DUBLIN 16
DUBLIN 17
DUBLIN 18
DUBLIN 20
DUBLIN 22
DUBLIN 24
```

If they choose any of the Dublin Postal regions. We again have the problem that they can't check out. Because it's not a valid county. WooCommerce doesn't support Dublin Postal regions, so all those should be mapped to just county Dublin.

And lastly, none of the Irish Address fields in PayPal are mandatory. So you can use an address without any county. Which results in the following error when you get back to WooCommerce: `Billing County is a required field.`
Not sure how to solve that one. It probably won't happen much, since most people will select a county. But it is possible that it will happen.

To be honest I've been living in Ireland now for 7 years and I'm still surprised at how the postal system works, but it is very reliable. In a lot of places the postman still knows everyone by their first name. Especially in rural areas, houses have names instead of numbers. There are no street names, but instead they just have a name which covers a certain area. They introduced postcodes a few years ago which is supposed to make it easier to structure the addresses, but it's still not in full use. Which is why it's still not a required field.

PR #332, opened 2017-11-14
Updates for the original PR

bor0:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/331#issuecomment-344204079.


bor0:
Hey @mikkamp, can you check this one? Thanks!

mikkamp:
@bor0 I just tested those two scenario's. Both a Dublin postal district and an address without a county works now. I'm sure the Irish PayPal Express users will love you for that.

Issue #340, opened 2017-11-28
Percentage Coupon usage causes PayPal error 10413 : The totals of the cart item amounts do not match order amounts


gateway-davidtouchette:
Usage of a percent off coupon causes discrepancy in cart total and user cannot checkout!  We have had to disable PP checkout while our holiday coupons are active.

I have tried updating our staging env to latest WooCommerce (v3.2.5) + WooCommerce PayPal Express Checkout Gateway (v1.4.5) but the issue is exactly the same.

Probable duplicate of #335 

### PROD Details
WordPress v4.8.3
Divi Theme v3.0.41
WooCommerce v3.0.4
Coupon definition: Percentage Discount: 40% all items in shop, unlimited usage, no restrictions

### Reproduction steps
Add coupon as defined above
Add item to cart
Apply coupon code
Click express paypal checkout button
Receive error message

### Logs from STAGING test
11-27-2017 @ 16:17:32 - WC_Gateway_PPEC_Client::_request: remote request to https://api-3t.sandbox.paypal.com/nvp with params: Array
(
    [LOGOIMG] => 
    [HDRIMG] => 
    [PAGESTYLE] => 
    [BRANDNAME] => To Love, Honor and Vacuum
    [RETURNURL] => https://sheilagregoire.staging.wpengine.com/checkout/?woo-paypal-return=true
    [CANCELURL] => https://sheilagregoire.staging.wpengine.com/cart/?woo-paypal-cancel=true
    [LANDINGPAGE] => Login
    [SOLUTIONTYPE] => Sole
    [PAYMENTREQUEST_0_PAYMENTACTION] => sale
    [PAYMENTREQUEST_0_INSURANCEAMT] => 0
    [PAYMENTREQUEST_0_HANDLINGAMT] => 0
    [PAYMENTREQUEST_0_CUSTOM] => 
    [PAYMENTREQUEST_0_INVNUM] => 
    [PAYMENTREQUEST_0_CURRENCYCODE] => USD
    [PAYMENTREQUEST_0_AMT] => 2.99
    [PAYMENTREQUEST_0_ITEMAMT] => 2.99
    [PAYMENTREQUEST_0_SHIPPINGAMT] => 0
    [PAYMENTREQUEST_0_TAXAMT] => 0
    [PAYMENTREQUEST_0_SHIPDISCAMT] => 0
    [NOSHIPPING] => 1
[...]

PR #341, opened 2017-11-28
Fix paypal error 10413

DragosMocrii:
This fixes Issue #340. Code changed from **line_subtotal** to **line_total**, because line_total accounts for the amount changes (such as coupon discounts), whereas **line_subtotal** doesn't.


breadadams:
I've tested this and it fixes the issue we're currently having with `master` where the PP Express popup closes after loading for a second or two.

However once in PayPal, there's no indication of the discount & original price:
image
(Price: 1595, Discount: 319, Total: 1276)

Not sure if that's something you can include?

bor0:
Thank you for the contribution @DragosMocrii. Thanks for testing, @breadadams!

PR #345, opened 2017-12-04
Include discount in line items (if added)

### LINKED ISSUE #344: No mention of discount in PayPal window
After the recent fix in #341, I [mentioned](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/341#issuecomment-348480996) that the details weren't showing on the PP order:

PP Express order details

breadadams:
Close #344 

After PR:
image

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -480,7 +480,7 @@ class WC_Gateway_PPEC_Client {
 
 		$items = array();
 		foreach ( WC()->cart->cart_contents as $cart_item_key => $values ) {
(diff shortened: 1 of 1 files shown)


breadadams:
I've remove the description from the discount item (that's what was creating the _More info_ button), as this doesn't seem necessary. Ideally that would be the coupon names I guess (something to add to roadmap @bor0 💡 ). Result:

image

breadadams:
Actually something needs addressing here, it'll break if the discount is larger than the total 🤦‍♂️

jeffstieler:
Thanks for the contribution @breadadams! I'm sorry this sat for so long.

PR #346, opened 2017-12-06
Fix issue by re-using packages in case we don't have PP context avail…

### LINKED ISSUE #330: Live only - shipping address is empty
This is only occurring in live mode for some reason, so you'll need at least 2 live accounts for testing as PayPal will not allow merchants to purchase from themselves.  

When checking out with PayPal Express from the cart page, the problem occurs after returning to WooCommerce to confirm payment.  The shipping address is correctly passed through from PayPal and requests are sent to shipping methods like USPS and FedEx and rates are returned.  Then, a second request is sent to each shipping method that doesn't include the shipping address.  So, ultimately, no shipping methods can be found.  Both extensions return destination zip code errors such as `Missing value for ZipDestination.`.  

The shipping extensions used in my testing are: 

WooCommerce USPS Shipping
WooCommerce FedEx Shipping

@LaurenaRehbein @stevetested

bor0:
…able

Fixes #330.

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -845,6 +845,7 @@ class WC_Gateway_PPEC_Checkout_Handler {
 
 		update_post_meta( $old_wc ? $order->id : $order->get_id(), '_paypal_status', strtolower( $payment->payment_status ) );
 		update_post_meta( $old_wc ? $order->id : $order->get_id(), '_transaction_id', $payment->transaction_id );
+		unset( WC()->session->ppec_shipping_info );
 
 		// Handle $payment response
 		if ( 'completed' === strtolower( $payment->payment_status ) ) {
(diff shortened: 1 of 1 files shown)


bor0:
To put a little background on this PR, based on the investigation that we did with @gedex:

By bisect, we found that this was caused by https://github.com/woocommerce/woocommerce/commit/2ff1ee145e5a3cce31e6c34190cc667056452edf. But that commit by itself makes total sense. What was caused by that commit is that now it (properly) triggers the `update_order_review` based on Geolocation, but as a result it revealed a bug in how hooks for shipping are handled.

I think the conclusion here is, for extension developers, when hooking on `woocommerce_cart_shipping_packages` make sure they return the same packages consistently and not depend on `$_GET`, since in the AJAX call for `update_order_review` the `$_GET` context may not be available. And when you return different shipping packages (hash mismatch) then the `get_rates_for_package` call within `WC_Shipping::calculate_shipping_for_package` will re-trigger shipping extensions.

gedex:
@bor0 Make sure to link this commit as well https://github.com/woocommerce/woocommerce/commit/75cc4f91b534ce3114d19da80586bacd083bb5a8 so we've a context why this change is needed.

PR #347, opened 2017-12-06
Add hooks for request endpoint and settings

dechov:
Adds filter `woocommerce_paypal_express_checkout_request_endpoint` (alongside the existing `woocommerce_paypal_express_checkout_request_body`) to be able to relay API requests via another server. This will be used for authenticating API requests on the WCS server. (Context in p7bbVw-2kH-p2.)

Also adds a settings filter — see [precedent in Stripe plugin](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6d8164689007384cd7178d2d65bdec3a1c2094bb/includes/settings-stripe.php#L6) — in order to be able to disable settings we know won't work with an API Subject lacking non-default permissions.

Let me know if MITM or other security concerns come to mind, though I don't believe this would make anything possible that isn't already.

<s>Marking "in progress" until https://github.com/Automattic/woocommerce-services/pull/1254 is approved without further changes to this plugin necessary.</s>

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -96,11 +96,12 @@ class WC_Gateway_PPEC_Client {
 	 * @return string
 	 */
 	public function get_endpoint() {
-		return sprintf(
+		$endpoint = sprintf(
 			'https://%s%s.paypal.com/nvp',
 			$this->_credential->get_endpoint_subdomain(),
 			'sandbox' === $this->_environment ? '.sandbox' : ''
 		);
+		return apply_filters( 'woocommerce_paypal_express_checkout_request_endpoint', $endpoint, $this->_credential->get_username() );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/class-wc-gateway-ppec-client.php
```diff
@@ -96,11 +96,11 @@ public function set_environment( $environment ) {
 	 * @return string
 	 */
 	public function get_endpoint() {
-		return sprintf(
+		return apply_filters( 'woocommerce_paypal_express_checkout_request_endpoint', sprintf(
```
bor0: Just wondering if it'd be useful to pass some context to this filter, say `$this->_environment`

dechov: 👍 I had that originally and removed it without considering that it would actually be generally useful.

dechov:
@bor0 The endpoint filter is no longer being used, since https://github.com/Automattic/woocommerce-services/pull/1254 is now using a different method of proxying the request. As a cleanup task, I'll go ahead and open another PR to revert the change that added it, unless you think it would be useful to leave it in.

PR #351, opened 2017-12-12
Add support for custom order IDs

SiR-DanieL:
In https://wordpress.org/support/topic/wc-order-numbers-not-going-through-to-paypal/ the user is trying to pass over their custom order ID, but there's no way to do so. I think the order ID should use the same filter used by WooCommerce.

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -794,13 +794,14 @@ class WC_Gateway_PPEC_Client {
 	 * @return array Params for DoExpressCheckoutPayment call
 	 */
 	public function get_do_express_checkout_params( array $args ) {
-		$settings  = wc_gateway_ppec()->settings;
-		$order     = wc_get_order( $args['order_id'] );
+		$settings     = wc_gateway_ppec()->settings;
+		$order        = wc_get_order( $args['order_id'] );
 
-		$old_wc    = version_compare( WC_VERSION, '3.0', '<' );
-		$order_id  = $old_wc ? $order->id : $order->get_id();
-		$details   = $this->_get_details_from_order( $order_id );
-		$order_key = $old_wc ? $order->order_key : $order->get_order_key();
+		$old_wc       = version_compare( WC_VERSION, '3.0', '<' );
+		$order_id     = $old_wc ? $order->id : $order->get_id();
+		$order_number = $old_wc ? (string) apply_filters( 'woocommerce_order_number', $order->get_id(), $order ) : $order->get_order_number();
+		$details      = $this->_get_details_from_order( $order_id );
+		$order_key    = $old_wc ? $order->order_key : $order->get_order_key();
 
 		$params = array(
 			'TOKEN'                          => $args['token'],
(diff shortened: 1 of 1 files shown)


includes/class-wc-gateway-ppec-client.php
```diff
@@ -817,7 +817,7 @@ public function get_do_express_checkout_params( array $args ) {
 			'PAYMENTREQUEST_0_PAYMENTACTION' => $settings->get_paymentaction(),
 			'PAYMENTREQUEST_0_INVNUM'        => $settings->invoice_prefix . $order->get_order_number(),
 			'PAYMENTREQUEST_0_CUSTOM'        => json_encode( array(
-				'order_id'  => $order_id,
+				'order_id'  => apply_filters( 'woocommerce_order_number', $order_id ),
```
bor0: If we want to pass the order number instead of the order id, why not just do `$order->get_order_number()` here?

I think this will cause issues, since we're reusing this custom variable in `WC_Gateway_PPEC_PayPal_Request_Handler::get_paypal_order`, where we get an order based on `order_id` (I assume that's the one we send here).

For example (assuming I have order ID 507):
```php
add_filter( 'woocommerce_order_number', function( $order_id ) {
	return 'Fancy order: ' . $order_id;
} );

add_action( 'init', function() {
	assert( false !== wc_get_order( 507 ) );
	assert( false !== wc_get_order( 'Fancy order: ' . 507 ) );
} );
```
Second assertion fails.

bor0: Maybe we can send another custom parameter to indicate the actual order id, and re-use that.

SiR-DanieL:
Oh yeah, I like that, but does PayPal support it? can I send anything in that field?

bor0:
I think we can send anything for the `CUSTOM` field

SiR-DanieL:
Ok, changed @bor0

includes/class-wc-gateway-ppec-client.php
```diff
@@ -794,13 +794,14 @@ public function do_express_checkout_payment( array $params ) {
...
+		$settings     = wc_gateway_ppec()->settings;
+		$order        = wc_get_order( $args['order_id'] );
 
-		$old_wc    = version_compare( WC_VERSION, '3.0', '<' );
-		$order_id  = $old_wc ? $order->id : $order->get_id();
-		$details   = $this->_get_details_from_order( $order_id );
-		$order_key = $old_wc ? $order->order_key : $order->get_order_key();
+		$old_wc       = version_compare( WC_VERSION, '3.0', '<' );
+		$order_id     = $old_wc ? $order->id : $order->get_id();
+		$order_number = $old_wc ? (string) apply_filters( 'woocommerce_order_number', $order->get_id(), $order ) : $order->get_order_number();
```
bor0: We can use `$order_number = $order->get_order_number();`, it's defined on both WC 2.6 and WC 3.x :)

bor0:
@SiR-DanieL thanks! Just one more minor thing and it's good to go

SiR-DanieL:
Fixed @bor0

bor0:
Thanks! This will be a part of the 1.5.0 release.

PR #352, opened 2017-12-12
Implement PayPal credit on checkout

### LINKED ISSUE #292: Offer PayPal Credit on Checkout
Ticket: 595437-zd-woothemes

Currently, plugin only offers PayPal Credit on the cart page, not the checkout. Because of this, customers who use PayPal Credit are not forced to sign up for an account. Even with guest checkout unchecked, customers are sent directly to PayPal Credit and not directed back to sign up for an account.

Please consider adding this feature so our users can take advantage of the PayPal credit option.

This feature is necessary when selling digital and membership based products.

bor0:
Fixes #292.

- Refactor checkout methods (code duplication)
- Implement a separate gateway class for PayPal credit (with a separate logo, title)
- Refactor some methods so that they can be used in the context of woocommerce filter

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -649,11 +649,12 @@ class WC_Gateway_PPEC_Checkout_Handler {
 	/**
 	 * Handler when buyer is checking out from checkout page.
 	 *
-	 * @param int $order_id Order ID
+	 * @param int  $order_id Order ID.
+	 * @param bool $use_ppc  Whether to use PayPal credit.
 	 *
 	 * @return string Redirect URL.
 	 */
-	public function start_checkout_from_checkout( $order_id ) {
+	public function start_checkout_from_checkout( $order_id, $use_ppc ) {
 		$settings     = wc_gateway_ppec()->settings;
 
 		$context_args = array(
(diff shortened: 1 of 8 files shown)


bor0:
Please let me know if you have any suggestions to the flow. Right now, PayPal Credit is treated as a separate gateway and is available to be selected in the checkout screen as an option (given PayPal Credit is available in the specific country and is enabled)

dwainm:
@bor0 It is essentially a different service so fine to have it's own "gateway" class. This makes it easier to fix in future. One thing I couldn't test was the checkout, maybe as my account/country doesn't allow for PP credit.

bor0:
@dwainm ah yea, sorry. It's only supported in the US (I had US set as my store location)

includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -598,74 +598,78 @@ protected function is_success( $response ) {
...
+
+	/**
+	 * Handler when buyer is checking out from checkout page.
 	 *
-	 * @throws PayPal_API_Exception
+	 * @param int  $order_id Order ID.
+	 * @param bool $use_ppc  Whether to use PayPal credit.
 	 *
-	 * @param int $order_id Order ID
+	 * @return string Redirect URL.
```
dwainm: Mayb some context on the use_ppc parameter?

PR #353, opened 2017-12-13
WC 3.3 compat

bor0:


includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -994,4 +998,20 @@ public function filter_wc_checkout_params( $params ) {
...
+	/**
+	 * Compatibility method for WC >= 3.3.
+	 *
+	 * @param array  $params
+	 * @param string $handle
+	 *
+	 * @return string URL
+	 */
+	public function filter_wc_checkout_params_post33( $params, $handle ) {
+		if ( 'wc-checkout' !== $handle ) {
```
dwainm: @bor0 we don't need 2 functions.

What if we add to the first function parameters.
`( $params, $handle = '' )`

Then we have a default so there will be no warning.

Removing this function also makes it easy to remove the second filter at a later stage.

includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -60,7 +60,7 @@ public function __construct() {
 	 */
 	public function init() {
 		if ( version_compare( WC_VERSION, '3.3', '<' ) ) {
-			add_filter( 'wc_checkout_params', array( $this, 'filter_wc_checkout_params' ), 10, 1 );
+			add_filter( 'wc_checkout_params', array( $this, 'filter_wc_checkout_params' ), 10, 2 );
 		} else {
 			add_filter( 'woocommerce_get_script_data', array( $this, 'filter_wc_checkout_params_post33' ), 10, 2 );
```
dwainm: @bor0 this filter still references the remove function.

bor0: Ah good catch man! Thanks! Updated in 41401dd.

PR #373, opened 2018-01-23
Revert adding request endpoint filter

dechov:
Reverts the change in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/347 that added an API endpoint URL filter, which didn't end up being used for the purposes that it was intended for.


dechov:
👍 The change that introduced it hasn't been released, so anyone who could be using it would have been knowingly working off unstable code in `master`.

Issue #376, opened 2018-01-30
403 Error when i try to set feature image for any post.


malsubrata:
If I call `WC()->payment_gateways()->payment_gateways` in `admin_init` hook then there is a 403 error in when i triad to set feature image for any post or post type.

Please help.

Thanks in advance!

malsubrata:
I have found the error!
I believe that we can't call `wp_enqueue_media()` function directly to `WC_Gateway_PPEC` class constructor.
That's why it show 403 error when ever I want to set feature image for any post

malsubrata:
Any update for me? I need to fix this please help me. You can refer to [Function Reference/wp enqueue media](https://codex.wordpress.org/Function_Reference/wp_enqueue_media) for more details.
Thank you.

malsubrata:
@mikejolley Please have a look at this issue because it causes conflict with other plugins. Looking forward to your co-operation regarding this.
Thank you.

malsubrata:
@mikejolley it's been 25 day's I have created this thread but there is nobody to look into this issue really upsets!

mikejolley:
cc @laurendavissmith

imankiyanoosh:
Hello everyone I have the same problem but when I try to select an image as you can see in the attachment there is a file selected which does not have name or link. I try to upload different images but none of them are working
![Screenshot (398)](https://user-images.githubusercontent.com/58703576/102422818-6d97d280-3fbc-11eb-94ef-ff3a9ec7ceeb.png)

PR #393, opened 2018-02-24
Fix-issue-#376

malsubrata:
Issue #376 fixed please check!


roykho:
Already done here https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/392

PR #401, opened 2018-04-02
Pass woo return args to ajax

### LINKED ISSUE #326: When returning from PayPal, place order buttons says "proceed to payment"
From the cart page, if you choose to checkout with PayPal Express, and successfully authorize the payment, on return to the store to finalize payment, the "place order" buttons says "proceed to payment", which is confusing. It should say "Place order".

Looking deeper into the code, the actual button has no value in the HTML, and is set with JavaScript.

Tested with WordPress 4.8.2, Woocommerce 3.2.1, PayPal Express Checkout Gateway 1.4.4.

bor0:
Fixes #326.

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -67,6 +67,23 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 		} else {
 			add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
 		}
+
+		add_filter( 'woocommerce_ajax_get_endpoint', array( $this, 'pass_return_args_to_ajax' ), 10, 2 );
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


includes/abstracts/abstract-wc-gateway-ppec.php
```diff
@@ -67,6 +67,23 @@ public function __construct() {
...
+	/**
+	 * Pass woo return args to AJAX endpoint when the checkout updates from the frontend
+	 * so that the order button gets set correctly.
+	 *
+	 * @param  string $request Optional.
+	 * @return string
+	 */
+	public function pass_return_args_to_ajax( $request ) {
+		if ( isset( $_GET['woo-paypal-return'] ) ) {
+			$request .= '&woo-paypal-return=1';
```
bor0: Can't use `add_query_arg` on `$request` here because it will automatically encode the percents in `wc-ajax=%%endpoint%%`.

PR #406, opened 2018-04-09
Make billing address not require by default and filterable. Closes #386

### LINKED ISSUE #386: Billing address is not required even if PayPal Express is disabled
PayPal Express checkout has a feature where the Billing fields are set to non-required fields. This works great when this needs to be used for PayPal Express. However this is always turned on as soon as the plugin is active. Even if we disable the PayPal Express as a checkout option, it still modifies the billing fields.

Should we add an additional check there to make sure the checkout option is actually enabled?

roykho:
* Fix - Prevent filtering if PPEC is not enabled.
* Tweak - Default billing address to be required.
* Add - Hook to make billing address not required `woocommerce_paypal_express_checkout_address_not_required` (bool).

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -114,11 +114,20 @@ class WC_Gateway_PPEC_Checkout_Handler {
 	 * See also filter_billing_fields below.
 	 *
 	 * @since 1.2.1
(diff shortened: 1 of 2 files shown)


roykho:
Yes it does. There is no way around that.

bor0:
I just checked this and that's correct. The only way is to hook into `checkout.js` and do the filtering there since there is no AJAX fired when we change a gateway.

PR #419, opened 2018-06-05
Set requiredness of billing fields based on whether shipping is neede…

### LINKED ISSUE #417: Urgent: paypal-express-checkout fails to place an order: several billing fields are missing
Here's the error message after returning from paypal.
image

Here's a video explanation
https://youtu.be/y0MICr4f_1c

```
Billing Street address is a required field.
Billing Town / City is a required field.
Billing Province is a required field.
Billing Postcode / ZIP is a required field.
```

bor0:
…d or not

Fixes #417.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Changelog ***
 
+= 1.5.6 - 2018-xx-xx =
+* Fix    - Virtual products cause issues with billing details validation.
+
 = 1.5.5 - 2018-05-23 =
 * Update - WC 3.4 compatibility
 * Update - Privacy policy notification.
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -124,7 +124,7 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			return $fields;
 		}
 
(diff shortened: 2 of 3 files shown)


tommyshellberg:
This works for me but the billing address is basically empty, I'm assuming its the best that we can get in this scenario?

https://cld.wthms.co/HfacHY

bor0:
Yes, I believe so. See #286 as a related ticket.

tommyshellberg:
👍

bor0:
@roykho when we send `NOSHIPPING` (in case of virtual products, see ticket above) to PP it doesn't give back any addresses to attach to.

In this case we must unset address details to not be required.

bor0:
If you use that code and filter is false by default it will always evaluate to false, regardless of `NOSHIPPING`

roykho:
I am not 100% sure about this, but will take your word for it. :)

bor0:
@tommyshellberg it looks good to me but can you please also confirm that this fix does not break #386?

tommyshellberg:
@bor0 it looks okay to me, I'm able to check out with a virtual product, and the Billing fields are still required.

bor0:
Thank you for taking the time to verify. I will do a release with this fix included.

Issue #418, opened 2018-06-01
Image settings description tooltip is misaligned


dechov:
The custom image settings fields' tooltip shows up in one of multiple wrong places:

<img width="354" alt="screen shot 2018-06-01 at 9 38 47 am" src="https://user-images.githubusercontent.com/1867547/40843672-c21059be-657f-11e8-86a0-8e22262929ab.png">

<img width="660" alt="dmfnhrrgt9-3000x3000" src="https://user-images.githubusercontent.com/1867547/40843676-c49f56d0-657f-11e8-8747-e3208e514541.png">
(thanks @kellychoffman)

Should appear directly left of the "Add image" button, aligned with "Brand Name" and the rest of the settings.

PR #427, opened 2018-06-14
Adjust markup to fix help tip alignment for image settings

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/418

Moves tooltip into `<label>` to catch up with change in WC core help tip styling: https://github.com/woocommerce/woocommerce/commit/d8814f93819a41a64f9f6571542488de8db65dab

Before | After
-- | --
screen shot 2018-06-14 at 2 25 42 pm | screen shot 2018-06-14 at 2 26 07 pm

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -532,8 +532,7 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 		?>
 		<tr valign="top">
 			<th scope="row" class="titledesc">
-				<?php echo $this->get_tooltip_html( $data ); ?>
-				<label for="<?php echo esc_attr( $field_key ); ?>"><?php echo wp_kses_post( $data['title'] ); ?></label>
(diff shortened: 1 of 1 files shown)


includes/abstracts/abstract-wc-gateway-ppec.php
```diff
@@ -532,8 +532,7 @@ public function generate_image_html( $key, $data ) {
 		?>
 		<tr valign="top">
 			<th scope="row" class="titledesc">
-				<?php echo $this->get_tooltip_html( $data ); ?>
-				<label for="<?php echo esc_attr( $field_key ); ?>"><?php echo wp_kses_post( $data['title'] ); ?></label>
+				<label for="<?php echo esc_attr( $field_key ); ?>"><?php echo wp_kses_post( $data['title'] ); ?> <?php echo $this->get_tooltip_html( $data ); ?></label>
```
jeffstieler: Nitpick: does the tooltip require a leading space? Also, you can combine the `echo` statements with commas.

dechov: Good feedback – I think I'll leave it as it is, to align with https://github.com/woocommerce/woocommerce/commit/d8814f93819a41a64f9f6571542488de8db65dab.

PR #437, opened 2018-06-29
Fix PHP warning due to incorrect assumption of setting value type

dechov:
Appears when PayPal Credit is not supported and Hide Funding Method(s) multiselect setting is empty, the value for which is encoded as `''` rather than `array()`.

To reproduce issue, select country other than US or currency other than USD so that PayPal Credit is not supported. Save empty selection for "Hide Funding Method(s)" with vertical layout, and observe PHP warning when rendering button on the front-end.

Reported in https://wordpress.org/support/topic/get-warning-when-enable-credit-card-in-express-smart-button/

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -277,7 +277,9 @@ class WC_Gateway_PPEC_Cart_Handler {
 
 		if ( ! wc_gateway_ppec_is_credit_supported() ) {
 			$data['credit_enabled'] = 'no';
-			if ( ! in_array( 'CREDIT', $data['hide_funding_methods'] ) ) {
+			if ( ! $data['hide_funding_methods'] ) {
+				$data['hide_funding_methods'] = array( 'CREDIT' );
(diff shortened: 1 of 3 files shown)


includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -277,7 +277,9 @@ protected function get_button_settings( $settings, $context = '' ) {
 
 		if ( ! wc_gateway_ppec_is_credit_supported() ) {
 			$data['credit_enabled'] = 'no';
-			if ( ! in_array( 'CREDIT', $data['hide_funding_methods'] ) ) {
+			if ( ! $data['hide_funding_methods'] ) {
+				$data['hide_funding_methods'] = array( 'CREDIT' );
+			} elseif ( ! in_array( 'CREDIT', $data['hide_funding_methods'] ) ) {
```
bor0: Maybe safer to do `is_array( $x ) && ! in_array( 'CREDIT', $x )`?

bor0: Do we want an additional `is_array` check here?

Issue #439, opened 2018-07-02
No longer supports Automatic Recurring Payments when making any changes


jrick1229:
## Issue

When using PPEC as a gateway for Subscriptions, it is no longer considered as being able to accept Automatic Recurring Payments after making any changes to the settings.

Here's a screenshot of when first uploading PPEC: http://nimb.ws/HuKWI9 (using PPEC v1.6.0)

Here's a screenshot after adding my Sandbox Reference Transactions details: http://nimb.ws/vmloBs

The credentials are correct and the settings are how they should be, as they worked prior to updating to v1.6.0.

We've seen at least one customer with this issue and I'm unsure what could have changed that would cause this issue.

## Steps to Replicate

1. Upload WooCommerce v3.4.2
2. Upload PPEC v1.6.0
3. Upload Subscriptions v2.2.21
4. Notice that PPEC is initially showing 'Automatic Recurring Payments' as accepted
5. Fill in your details for login to PayPal
6. Notice that PPEC is now not compatible with 'Automatic Recurring Payments'
7. Make as many changes as you like, but it will no longer change back

Issue #436, opened 2018-06-29
Mandate Smart Payment Buttons and remove embedded image links


dechov:
https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409 added "Smart Payment Buttons" as an optional mode, refactored some code, but deleted nothing. Once we are confident that both new and existing merchants should have no serious issues using Smart Payment Buttons, we can make it the only option, as anticipated in the messaging to existing merchants:

<img width="1019" alt="screen shot 2018-06-25 at 11 58 20 am" src="https://user-images.githubusercontent.com/1867547/42113582-b623d48c-7bb9-11e8-93bc-f5e3915a86f4.png">

Provided that redirection to PayPal is unnecessary to keep as a fallback, then we can remove:
- legacy implementation of image links and the original JS integration on the cart page
- the conditional logic for switching between modes introduced in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409
- back-end code supporting redirection to PayPal

PR #442, opened 2018-07-03
Modify Simple Payment Button gateway to inherit from existing gateway classes

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/439 (and subscriptions not working in SPB in general).

Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/444

Instead of extending the base abstract class, SPB gateway now extends `WC_Gateway_PPEC_With_PayPal`. **Duplicates** `WC_Gateway_PPEC_With_SPB` code into a separate `WC_Gateway_PPEC_With_SPB_Addons` class (avoiding use of traits to support PHP <5.4), which will no longer be necessary after https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/436.

### DIFF
--- a/includes/class-wc-gateway-ppec-gateway-loader.php
+++ b/includes/class-wc-gateway-ppec-gateway-loader.php
@@ -22,6 +22,7 @@ class WC_Gateway_PPEC_Gateway_Loader {
 		require_once( $includes_path . 'class-wc-gateway-ppec-with-paypal-credit.php' );
 		require_once( $includes_path . 'class-wc-gateway-ppec-with-paypal-addons.php' );
 		require_once( $includes_path . 'class-wc-gateway-ppec-with-spb.php' );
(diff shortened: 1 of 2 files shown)


bor0:
I don't understand why we need to duplicate the exact same class?

dechov:
I would prefer not to – is there a better way to inherit from one or the
other original class? Need to leave the original classes alone for now, but
extend each in the exact same way.

bor0:
Why do we need 2 classes? :) Is one of them planned for additional implementation in the future?

dechov:
One class extends `WC_Gateway_PPEC_With_PayPal`, while the other extends `WC_Gateway_PPEC_With_PayPal_Addons` (that being the only difference – I may have made it seem like there was no difference). This PR reintroduces the functionality from those two classes – extending them with the SPB changes – and and the condition that chooses between them (i.e. whether subscriptions are available).

Issue #438, opened 2018-07-02
Smart Payment Buttons not working on mobile product pages


bondimedical3:
Steps to reproduce: 
Notes: Bug is only present on mobiles

1. Clear cache in mobile browser 
2. Navigate to a product page with a Smart Payment Button 
3. Click button

What happens:
A second screen opens and begins to load but then abruptly closes.

Expected result: 
Second screen opens redirecting customer to PayPal

Other Notes:
Smart buttons on Cart and Checkout pages of mobile site work correctly.

bondimedical3:
Hi @dechov - I updated to 1.6.1 but the issue is still present. The PayPal page starts to load and then abruptly closes.

bondimedical3:
You are correct in that this bug is not restricted to mobiles but for cases where no session has been created. I tested the desktop version of my website in a incognito window and the PayPal window does not open.

dechov:
@bondimedical3 Thank you for following up. https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/455 is meant to take care of this once and for all. Can you confirm whether 1.6.3 fixes the problem for you?

bondimedical3:
Hello Sir. Yes, it appears to be working correctly now. Thank you for your help.

dechov:
Thanks for verifying @bondimedical3 

Fixed by https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/455

PR #445, opened 2018-07-04
Fix single product Smart Payment Button failing without existing session

dechov:
The nonce created with no session will not match the nonce verified when a session is created on AJAX POST. This causes AJAX requests to fail, breaking cart generation and SPB payment from single product pages.

This PR makes sure a session is there by the time the nonce is created. Should address reports of PayPal window closing immediately (https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/438, [forum thread](https://wordpress.org/support/topic/many-bugs-in-1-6-0-version/)).

To reproduce broken payments with Smart Payment Buttons enabled, open single product page in a new session (e.g. Firefox Private Browsing), click the PayPal Checkout button, and see the PayPal window open and immediately close.

To reproduce failed cart generation (with or without Smart Payment Buttons), open single product page in a new session, enter a quantity greater than 1, click the PayPal Checkout button, and observe only 1 item being listed.

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -331,7 +331,7 @@ class WC_Gateway_PPEC_Cart_Handler {
 				'page'                 => $page,
 				'button_color'         => $settings->button_color,
 				'button_shape'         => $settings->button_shape,
-				'start_checkout_nonce' => wp_create_nonce( '_wc_ppec_start_checkout_nonce' ),
+				'start_checkout_nonce' => $this->create_nonce( '_wc_ppec_start_checkout_nonce' ),
 				'start_checkout_url'   => WC_AJAX::get_endpoint( 'wc_ppec_start_checkout' ),
 			);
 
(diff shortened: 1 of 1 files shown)


includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -362,13 +362,24 @@ public function enqueue_scripts() {
...
 				)
 			);
 		}
 	}
 
+	/**
+	 * Ensures a session is active, so that nonce is not invalidated by new session created on AJAX POST request.
+	 */
+	protected function create_nonce( $action ) {
+		if ( ! WC()->session->has_session() ) {
```
bor0: Do we need to check if `WC()->session` is populated?

dechov: I don't see it being checked in other places – as far as I understand, it should always be there for front-end requests:
https://github.com/woocommerce/woocommerce/blob/7b56840f360158b3f4a4f4ff924339fd70b82c17/includes/class-woocommerce.php#L476

bondimedical3:
I reported the same bug a few days ago.

PR #446, opened 2018-07-04
Add a filter to paypal locale

pierrebuet:
Paypal locale can be different from the site locale, as for example with de_DE_formal where paypal would need de_DE. The filtering of get_locale() being pretty cumbersome because of tons of options, adding a filter seems to be the easiest way.

### DIFF
--- a/includes/class-wc-gateway-ppec-settings.php
+++ b/includes/class-wc-gateway-ppec-settings.php
@@ -306,7 +306,7 @@ class WC_Gateway_PPEC_Settings {
 		if ( ! in_array( $locale, $this->_supported_locales ) ) {
 			$locale = 'en_US';
 		}
-		return $locale;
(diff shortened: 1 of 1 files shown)


includes/class-wc-gateway-ppec-settings.php
```diff
@@ -306,7 +306,7 @@ public function get_paypal_locale() {
 		if ( ! in_array( $locale, $this->_supported_locales ) ) {
 			$locale = 'en_US';
 		}
-		return $locale;
+		return apply_filters( 'woocommerce_paypal_express_checkout_get_paypal_locale', $locale );
```
dechov: I believe `get` is unnecessary in the name of the filter here – should be enough to filter the `paypal_locale` or just `locale` (with `woocommerce_paypal_express_checkout` as prefix).

pierrebuet: Good point!

PR #447, opened 2018-07-04
When cart is empty, don't show mini-cart buttons or attempt to render cart buttons

dechov:
Fixes buttons showing up in the mini-cart when the cart is empty:
screen shot 2018-06-27 at 4 50 51 pm

Also fixes a JS error any time an empty cart page is loaded.

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -301,7 +301,7 @@ class WC_Gateway_PPEC_Cart_Handler {
 
 		wp_enqueue_style( 'wc-gateway-ppec-frontend-cart', wc_gateway_ppec()->plugin_url . 'assets/css/wc-gateway-ppec-frontend-cart.css' );
 
-		$is_cart     = is_cart() && 'yes' === $settings->cart_checkout_enabled;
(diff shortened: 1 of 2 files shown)


includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -301,7 +301,7 @@ public function enqueue_scripts() {
 
 		wp_enqueue_style( 'wc-gateway-ppec-frontend-cart', wc_gateway_ppec()->plugin_url . 'assets/css/wc-gateway-ppec-frontend-cart.css' );
 
-		$is_cart     = is_cart() && 'yes' === $settings->cart_checkout_enabled;
+		$is_cart     = is_cart() && ! WC()->cart->is_empty() && 'yes' === $settings->cart_checkout_enabled;
```
bor0: Same note on previous PR, not sure if there's a case where the cart may not be populated.

dechov: Doesn't seem so, as it's used in other places, and is initialized on every front-end request: https://github.com/woocommerce/woocommerce/blob/7b56840f360158b3f4a4f4ff924339fd70b82c17/includes/class-woocommerce.php#L479

Issue #433, opened 2018-06-27
Tax not applied on the (Confirm your PayPal order) page at the checkout


madeincosmos:
#### Affected tickets

1243199-zen

#### What I expected

For the checkout to display accurate amount of what the customer is going to pay before they place the order.

#### What happens instead

After returning from PayPal, tax is not applied on the (Confirm your PayPal order) page at all. Only after someone places an order they can see how much they've actually paid.

#### Steps to reproduce

1. Set prices exclusive of tax, tax based on Customer shipping address,
2. Add the following tax rate:

![Screenshot](http://cld.wthms.co/TZ3Ht+)
Screenshot: http://cld.wthms.co/TZ3Ht

3. Create any taxable product, add it to cart, and click the PayPal button on the cart page,
4. Login with PayPal and set a shipping address in California on the PayPal account. You can use the following one as a test:

```
test buyer
1 Main st
San Jose, CA 95131
United States (US)
```

5. Observe there are no taxes applied on the confirmation page even though the shipping address is in the US:

![Screenshot](http://cld.wthms.co/SI1CKT+)
Screenshot: http://cld.wthms.co/SI1CKT

6. Place the order and observe the tax got eventually charged:

![Screenshot](http://cld.wthms.co/Oj8cP9+)
Screenshot: http://cld.wthms.co/Oj8cP9

c-shultz:
I was able to reproduce this.

One thing to note is that, if the customer provides an address for shipping calculation, then the tax will be added and included in confirmation:

<img width="411" alt="image" src="https://user-images.githubusercontent.com/6209518/42398032-0fb3ddd6-811d-11e8-9714-7e439b5b71f3.png">

Otherwise, it looks like the address is not being grabbed from PayPal and included in confirmation. I'll be on janitorial duties next week as well, so I'll work on stamping this out then.

allendav:
@c-shultz - looks like this was fixed. Closing it. If I shouldn't have, please re-open.

PR #453, opened 2018-07-10
Fix for issue #433

c-shultz:
Addresses issues #433 

Sets current customer shipping and billing addresses from PayPal session as soon as the PayPal pop-up returns. This is necessary because the PPEC extension allows the customer to select an address in the PayPal pop-up, so the address must be retrieved from there:

image

Previously, even though the correct shipping address was displayed on the "Confirm your PayPal order" screen, it was not being used by WooCommerce to calculate the taxes. The correct taxes were only calculated at final checkout, so the final charged amount would not always match the confirmation screen.

Now the taxes are calculated from the same address that is displayed on the confirmation screen:
image

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -438,6 +438,9 @@ class WC_Gateway_PPEC_Checkout_Handler {
 		$session->payer_id           = $payer_id;
 		$session->token              = $token;
 
+		// Update customer addresses here from PayPal selection so they can be used to calculate local taxes.
+		$this->update_customer_addresses_from_paypal( $token );
+
 		WC()->session->set( 'paypal', $session );
 
 		try {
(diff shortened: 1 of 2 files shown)


includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -476,6 +479,48 @@ public function maybe_return_from_paypal() {
...
+		} catch ( PayPal_API_Exception $e ) {
+			wc_add_notice( $e->getMessage(), 'error' );
+			return;
+		}
+		$shipping_details = $this->get_mapped_shipping_address( $checkout_details );
+		$billing_details  = $this->get_mapped_billing_address( $checkout_details );
+
+		$customer = WC()->customer;
+
+		// Update billing/shipping addresses.
```
laurendavissmith: Not needed for this PR but just for a future thought, I wonder if a method that wraps this functionality could DRY up this code quite a bit

budzanowski: @c-shultz @laurendavissmith have grep-ed the code and found no other place that was using this so I thought that it is OK for one time. When I think about it now I think that I would have explored using `set_props`  from `WC_Data` https://docs.woocommerce.com/wp-content/images/wc-apidocs/source-class-WC_Data.html#589-622 
`WC_Customer` is a subclass of `WC_Data`. `set_props` uses a nice reflection trick that makes the `set_props` and this code equivalent.

laurendavissmith: @budzanowski clever, i like that even better

c-shultz: That's a good idea!

@budzanowski, something along the lines of how it's used here? - https://github.com/woocommerce/woocommerce/blob/3fa3830b6ceeaafda7eab8f5fa3534026913bd71/includes/class-wc-ajax.php#L293

That way, we can avoid as much repetition and the multiple "set" calls.

budzanowski: I was thinking about something along the lines of:

```php
	private function update_customer_addresses_from_paypal( $token ) {
		// Get the buyer details from PayPal.
		try {
			$checkout_details = $this->get_checkout_details( $token );
		} catch ( PayPal_API_Exception $e ) {
			wc_add_notice( $e->getMessage(), 'error' );
			return;
		}
		$shipping = $this->get_mapped_shipping_address( $checkout_details );
		$billing  = $this->get_mapped_billing_address( $checkout_details );
		$props = [];

		foreach(['address_1', 'address_2', 'city', 'postcode', 'state', 'country'] as $key ) {
			$props['shipping_' . $key] = $shipping[ $key ];
			$props['billing_' . $key]  = $billing[ $key ];
		}

		WC()->customer->set_props( $props );
	}
```

budzanowski: I don't have a very strong opinion that this is the best way. I just think that it is nicely short :)

PR #455, opened 2018-07-12
Address invalid nonce due to no customer session in a more robust way

dechov:
Ensures a customer session has been created early enough (for a consistent AJAX nonce) – before scripts are enqueued, at which point headers may have already been sent.

Partially reverts e5c5c47, which fixed the bug on some sites by initializing the session, but on other sites, the header template was being printed before the session could be saved, which was too late for saving the cookie, and so there was a new session (and therefore new nonce) when the AJAX request was processed.

(copied from https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/445:)

Testing the failure case on sites that still have this issue, such as https://jurassic.ninja/ sites:
> To reproduce broken payments with Smart Payment Buttons enabled, open single product page in a new session (e.g. Firefox Private Browsing), click the PayPal Checkout button, and see the PayPal window open and immediately close.

Again...
> Should address reports of PayPal window closing immediately (#438, [forum thread](https://wordpress.org/support/topic/many-bugs-in-1-6-0-version/)).


dechov:
Should fix https://wordpress.org/support/topic/cookie-cannot-be-set-3/

PR #458, opened 2018-07-23
Initial code for fixing issue with billing fields filters

### LINKED ISSUE #426: Version 1.5.6 filters all address fields to "optional"
**Since 1.5.6, if "require billing address" setting is unchecked, it filters all address fields to "optional" even if the customer selects a different gateway. If the setting *is* checked, it seems to have no effect.**

In `class-wc-gateway-ppec-checkout-handler.php` the function `filter_default_address_fields()` seems to do this but it applies that filter to the fields *even if PPEC is not the selected gateway*. This is very heavy-handed because it makes all fields non-required, which causes validation errors for other payment processors. The issue can be prevented by `add_filter( 'woocommerce_paypal_express_checkout_address_not_required', '__return_false' );` but this shouldn't be necessary.

bor0:
Fixes #426.

### DIFF
--- a/assets/js/wc-gateway-ppec-checkout-handler.js
+++ b/assets/js/wc-gateway-ppec-checkout-handler.js
@@ -1,8 +1,9 @@
 jQuery(document).ready(function($) {
-    $( '#order_review' ).on('change', 'input[name=payment_method]', function() {
-    	if ( $('#payment_method_ppec_paypal').size() ) {
-    		$('body').trigger( 'update_checkout' );
-			// TODO: Update billing fields
-    	}
-    });
+	$( '#order_review' ).on( 'change', 'input[name=payment_method]', function() {
+		if ( $( '#payment_method_ppec_paypal' ).size() ) {
+			$( 'body' ).trigger( 'update_checkout' );
+
+			// TODO: Update billing fields on the UI
+		}
+	});
 });
--- a/assets/js/wc-gateway-ppec-checkout-handler.js
+++ b/assets/js/wc-gateway-ppec-checkout-handler.js
@@ -1,0 +1,8 @@
+jQuery(document).ready(function($) {
+    $( '#order_review' ).on('change', 'input[name=payment_method]', function() {
+    	if ( $('#payment_method_ppec_paypal').size() ) {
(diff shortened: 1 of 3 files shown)


assets/js/wc-gateway-ppec-checkout-handler.js
```diff
@@ -0,0 +1,8 @@
+jQuery(document).ready(function($) {
+    $( '#order_review' ).on('change', 'input[name=payment_method]', function() {
+    	if ( $('#payment_method_ppec_paypal').size() ) {
+    		$('body').trigger( 'update_checkout' );
+			// TODO: Update billing fields
```
bor0: @claudiulodro any ideas how I can achieve this?

claudiulodro: Are you trying to refresh the billing fields? Should be able to do that with something like this:

```
jQuery( document.body ).trigger(  
    'country_to_state_changing', 
    [ jQuery( '#billing_country' ).val(), jQuery( '.woocommerce-billing-fields' ) ] 
)
```

The `country_to_state_changing` action handles updating the form when changing countries. Triggering it with the current country should reset them. If that doesn't meet your needs you'll likely have to go with something custom, as WC probably doesn't have what you're looking for built in.

jessepearson:
@bor0 Is this a completed fix or a WIP?

bor0:
Sorry @jessepearson, it's still WIP.

jessepearson:
No worries, just making sure 👍

dphyled:
Hello @bor0 ! Still WIP here? I see assignment was removed on the issue. 
Thank you!

dechov:
@bor0 I see the JS was removed in the latest commit – did you conclude that it was infeasible to update the required-ness of the fields dynamically? Is there any other way forward that you had in mind, or were you thinking of responding to changed payment method as out of scope for this PR?

bor0:
> did you conclude that it was infeasible to update the required-ness of the fields dynamically

Yep. Related: https://a8c.slack.com/archives/C0E1AV8T0/p1532432033000259

bor0:
Closing stale PR. It still can be used as a reference if needed.

PR #461, opened 2018-07-31
Update readme.txt to README.MD

leongjiameng:
This can ease the reading.


allendav:
Hi @leongjiameng - sorry - we need readme.txt to be so named as that file is required by the WordPress.org repository

Issue #451, opened 2018-07-06
"Smart Payments" allows orders to go through without accepting Terms


zwiesman31:
If you enable the "Smart Payment buttons" option in the extension, you can make the payment even if you don't accept the Terms of Service.

![http://cld.wthms.co/JrkBnt](http://cld.wthms.co/JrkBnt+)
Direct link: http://cld.wthms.co/JrkBnt

Tested on a site with only the latest versions of [Storefront](https://woocommerce.com/products/storefront/), [WooCommerce](https://wordpress.org/plugins/woocommerce/) and [PayPal Checkout](https://woocommerce.com/products/woocommerce-gateway-paypal-checkout/) enabled.

This is only an issue when "Smart Payments" is enabled.

Ticket: 1264383-zen

yukikatayama:
1267580-zen reported here as well

dechov:
As far as I could tell, you weren't able to actually complete the payment – just the PayPal portion of it, after which it would show you an error – will be fixed in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/454. (Please correct me if you saw different behavior.)

TATIYA:
Hello
Same thing if a required field is empty you still can make a payment

bondimedical3:
Updated to 1.6.3 today and this issue is still present?

TATIYA:
Yes it still present

juliaamosova:
Hi @dechov pinging you here to see if you can provide an update here. Thanks!

madeincosmos:
Just tested this with 1.6.3 and indeed after confirming the payment with PayPal, I'm taken back to the checkout page with an error on the top:

![Screenshot](http://cld.wthms.co/IngRdo+)
Screenshot: http://cld.wthms.co/IngRdo

Still worth fixing the user flow though, cause it's quite confusing

dechov:
Hi @juliaamosova, @madeincosmos – sorry I had missed the earlier ping – we are working on this in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/475 and it should be merged and released soon.

dechov:
This fix is now in the latest release – thank you all for reporting and for your patience.

Issue #467, opened 2018-08-16
With require Product Addons, can checkout from product page without selecting the required addons


yukikatayama:
1350691-zen

Tested this on 2.1.3 and Product Addons 2.9.6

Whether it is with Smart Payments button or not, if you have a required field in Product Addons on product and if you do not select/enter info for the required field and checkout via product page, no error messaging shows:

![](http://cld.wthms.co/Iuiqha+)
Link to screenshot: http://cld.wthms.co/Iuiqha

Would expect to have similar behavior as if selecting Add to Cart button:

![](http://cld.wthms.co/PqxvWr+)
Link to screenshot: http://cld.wthms.co/PqxvWr

taoball:
We have experienced the same issue.  Customers can checkout without filling in product options.

PR #409, opened 2018-05-03
Add Smart Payment Buttons mode as alternative to directly embedded image links

dechov:
screen shot 2018-06-01 at 4 47 19 pm

(Toggled off by default for upgrading merchants, and on for new installs.)

Front-end view | Relevant settings | Fixes
-- | -- | --
Cart: screen shot 2018-06-01 at 9 57 57 am https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/99305a6524c65dc77f995febfd58dcfc18f4de7c | screen shot 2018-06-01 at 9 56 55 am https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/ea578d97abe0a2a41b3d7c910d932cdaf8e225c9 https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/2ca813a3256ccce9ec79a824b5b493992925750f https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/2318eb17eff01078a152a5ad734c7d821b6f975f https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/a9f4dec8b207e11548f377931ede72af0294190e
Single Product: screen shot 2018-06-01 at 9 10 31 am https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/d3727d6282b90a61c3aa578305a9406bbf9d619d | screen shot 2018-06-01 at 9 55 10 am https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/428ab6eb2a6ac34a6f624c91e1ae30b6622ff705 | https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/407
Checkout: screen shot 2018-06-01 at 10 06 08 am https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/baf4f69cd29bebac51d95bc5445559d88f0f1ddd | screen shot 2018-06-01 at 10 07 09 am https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/b1137aa36469ecceb15c8df0ce9bc698ddb37d87 | https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/403
[...]

PR #475, opened 2018-09-05
Validate checkout form via AJAX upon opening payment flow

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/451

Currently, the terms of service checkbox (and the rest of the checkout form's required fields) are validated only after the payment flow is complete, which causes an unnecessary loop in the flow and is too late. This change ensures that form data is sent along with the payment initialization call when a Smart Payment Button is clicked. On the Checkout screen, this triggers form validation on the server, and errors are reported.

Example with empty required fields:

screen shot 2018-09-05 at 6 45 31 pm

Example with terms of service unchecked:

screen shot 2018-09-05 at 6 45 49 pm

May be a step towards solving https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/467 (at least for the Smart Payment Buttons case).

### DIFF
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -2,20 +2,26 @@
 ;( function ( $, window, document ) {
 	'use strict';
 
-	var showError = function( errorMessage ) {
+	// Show error notice at top of checkout form, or else within button container
+	var showError = function( errorMessage, selector ) {
 		var $checkout_form = $( 'form.checkout' );
+
+		if ( ! $checkout_form || ! $checkout_form.length ) {
+			$( selector ).prepend( errorMessage );
+			return;
+		}
 
 		// Adapted from https://github.com/woocommerce/woocommerce/blob/ea9aa8cd59c9fa735460abf0ebcb97fa18f80d03/assets/js/frontend/checkout.js#L514-L529
 		$( '.woocommerce-NoticeGroup-checkout, .woocommerce-error, .woocommerce-message' ).remove();
 		$checkout_form.prepend( '<div class="woocommerce-NoticeGroup woocommerce-NoticeGroup-checkout">' + errorMessage + '</div>' );
 		$checkout_form.find( '.input-text, select, input:checkbox' ).trigger( 'validate' ).blur();
-		
+
 		var scrollElement = $( '.woocommerce-NoticeGroup-checkout' );
 		if ( ! scrollElement.length ) {
 			scrollElement = $checkout_form;
 		}
 		$.scroll_to_notices( scrollElement );
-		
+
 		$( document.body ).trigger( 'checkout_error' );
 	}
 
@@ -104,7 +110,7 @@
 								return '<li>' + message + '</li>';
 							} ).join( '' );
 
-							showError( '<ul class="woocommerce-error" role="alert">' + messageItems + '</ul>' );
+							showError( '<ul class="woocommerce-error" role="alert">' + messageItems + '</ul>', selector );
 							return null;
 						}
 						return response.data.token;
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -85,7 +85,7 @@
 						if ( ! response.success ) {
 							// Render error notice inside button container.
 							var $message = $( '<ul class="woocommerce-error" role="alert">' )
-								.append( $( '<li>' ).text( response.data.message ) );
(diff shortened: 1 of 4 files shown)


jeffstieler:
Functionally, this works as advertised - and the code looks nice and simple.

Some questions:
* Is it possible to keep the display of WC errors in the same location? Before this change, missing checkout field errors display at the top of the checkout form. After, they display above the PayPal button.
* Is it possible to avoid the PayPal modal window from being created when the form fails validation?

dechov:
- In order to put the errors on top (consistent with the core `submit` handling), I thought we'd have to duplicate [this code](https://github.com/woocommerce/woocommerce/blob/ea9aa8cd59c9fa735460abf0ebcb97fa18f80d03/assets/js/frontend/checkout.js#L514-L530), so I went with something simpler instead. Is there a way to access this function that I'm unaware of? I considered the possibility of spoofing the global `wc_checkout_params.checkout_url` and triggering `submit` before, but I don't see how you'd manipulate the handling of the response.

- Likewise, I'm not sure how we'd trigger anything other than PayPal's modal window (which opens immediately and waits for the `Promise` returned for `payment` to resolve) when the SPB is clicked.

jeffstieler:
> In order to put the errors on top (consistent with the core submit handling), I thought we'd have to duplicate this code, so I went with something simpler instead. Is there a way to access this function that I'm unaware of? I considered the possibility of spoofing the global wc_checkout_params.checkout_url and triggering submit before, but I don't see how you'd manipulate the handling of the response.

Ah, I see.. I think it would still be good to keep the core error message functionality even if it means duplicating some code. That function doesn't appear to be globally accessible.

> Likewise, I'm not sure how we'd trigger anything other than PayPal's modal window (which opens immediately and waits for the Promise returned for payment to resolve) when the SPB is clicked.

👍

dechov:
Moved the error notice to the top:

screen shot 2018-09-17 at 10 59 00 am

---

screen shot 2018-09-17 at 10 58 38 am

dechov:
The last commit broke error rendering in contexts other than the checkout screen, and I've just added it back in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/475/commits/49532e994ed4deb39d7c472846eb3cdb35d05651, same way as before:

screen shot 2018-09-18 at 8 59 03 am

I looked into rendering it at the top of the screen but it was rather messy to try and handle older WC versions where there was also less in common among different screens. For instance, this is the target of error notices for just the cart page:

- WC 2.6: `$( '.shop_table.cart' ).closest( 'form' )`
- WC 3.0-3.4: `$( '.woocommerce-cart-form' )`
- WC 3.5:  `$( '.woocommerce-notices-wrapper:first' ) || $( '.cart-empty' ).closest( '.woocommerce' ) || $( '.woocommerce-cart-form' )`

On the other hand, WC 3.5 will bring some consistency here (see [PR](https://github.com/woocommerce/woocommerce/pull/19794)) via a `.woocommerce-notices-wrapper` container – and also directly supporting output buffering the notice markup (like my original [attempt](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/409/commits/6cd19eee31cd573e0d6af15cc335d54f7c509d0e)). Though it will be some time until we can depend on this, perhaps we can check for a `.woocommerce-notices-wrapper` and fall back to the current way. Would that be an improvement?

jeffstieler:
@dechov how about we use the WC 3.5 logic, then fallback to the `selector` that's in scope. That seems like the easiest way to support 3.x.

dechov:
I've added a selector for the new container, keeping the `form.checkout` selector as a fallback:

&nbsp; | WC 3.5.0-beta.1 | WC 3.4.5
-- | -- | --
Single product | screen shot 2018-09-20 at 11 31 13 am | screen shot 2018-09-20 at 11 30 53 am
Cart | screen shot 2018-09-20 at 11 27 43 am | screen shot 2018-09-20 at 11 30 16 am
Checkout | screen shot 2018-09-20 at 11 31 26 am | screen shot 2018-09-20 at 11 32 06 am

Unfortunately I've just noticed that this approach for intercepting the validation error handling doesn't work at all for WC 2.6, which immediately calls `wc_add_notice` for each error, rather than first collecting them into an `$errors` array.

(Also, Smart Payment Buttons isn't working on the single product screen on 2.6.)

dechov:
Rebased to split first commit in two, isolating the change to send form data along with the AJAX call.

dechov:
Rebased on master – and also to split first commit in two, isolating the change to send form data along with the AJAX call.

bondimedical3:
Follow

Issue #474, opened 2018-08-31
Smart Payment Buttons does not populate PayPal Guest Checkout with billing info.


johndcoy:
Using PayPal Checkout the PayPal Guest Checkout is populated with billing information ( [#](http://cld.wthms.co/LUkEOm) ). When **Smart Payment Buttons** are enabled the data is not pulled from the checkout fields ( [#](http://cld.wthms.co/9fHwNV) ).

dechov:
Possible duplicate of https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/448 (if I understood that one correctly)

johndcoy:
Hey @dechov , That sounds right but I did confirm this is specific to using the Smart Payment Buttons. I can't confirm if that's the case with that issue.

jobthomas:
Another chat with this issue: 6414425-hc

Steps to reproduce:

* Use the smart button
* Choose an address associated with the PayPal account
* The result only returns a shipping and not a billing address

I get redirected to the shop page after choosing an address there:

![http://cld.wthms.co/i6SrDA](http://cld.wthms.co/i6SrDA+)
Link to image here: http://cld.wthms.co/i6SrDA.

(This is reproduced locally)

AmberSheppMZ:
This is happening with my website: 
When customers use paypal express to complete their purchase and add a different shipping address, once completed, woocommerce generates an order but changes the billing details to the shipping details too. Any news on when this will be fixed? Thanks!

dechov:
@jobthomas This appears to be a different issue – I can test a little later, but do you know if it is specific to Smart Payment Buttons, and if/how the [Require Billing Address](https://docs.woocommerce.com/document/paypal-express-checkout/#section-6) setting affects this?

@AmberSheppMZ Am I understanding correctly that the shipping address selected in the PayPal flow is ignored and the billing address from the PayPal flow is used instead?

jobthomas:
@dechov - ah, I thought that this was what @johndcoy  was describing. This is without the required billing address as that's something I'd need permission from PayPal for.

It's specific to the Smart button indeed. Just going through regular checkout doesn't make this happen (because then you'd add the checkout details there in the first place).

johndcoy:
Hey @jobthomas I saw a similar request for what you're referring to in the forums. Although the name and email show on the order confirmation page, the full shipping address is recorded after an order is placed. 

[...]

Issue #448, opened 2018-07-05
Address not being sent to PayPal 


pons1991:
When I am trying to pay with PayPal and choosing guest checkout, PayPal is asking for Billing address and my card details. Asking card details is fine. But I am bit concerned about asking Billing address. 

I dont want it to ask billing address. Please do the needful.

dechov:
I believe this is fixed by https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/476, which pre-fills the address form on the PayPal side with store checkout form values. This is included in the latest release.

Feel free to re-open and provide a bit more detail if this is still an issue – thank you!

PR #476, opened 2018-09-06
Provide customer address to PayPal flow even when no order is created

dechov:
Partially addresses https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/474
Should address https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/448 as well

If the customer has entered address data, this will be passed to the checkout form on the PayPal side to prefill the billing address fields (e.g. for the guest checkout flow when a credit card button is clicked).

In particular, this corrects the broken experience on the Checkout screen in which the (required) billing address is requested as part of the PayPal flow, without autofilled values – even after having been entered on the store's Checkout screen.

(Open to refactoring away code duplication, in particular [L681-L695](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/17f48123522c5a703ed0e2accd9c41f7103f9a5c/includes/class-wc-gateway-ppec-client.php#L681-L695) and [L736-L750](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/17f48123522c5a703ed0e2accd9c41f7103f9a5c/includes/class-wc-gateway-ppec-client.php#L736-L750).)

To test:
- With "Credit Card" not a hidden funding method in the settings
- And with the cart not empty
- Enter values into the billing address fields on the Checkout screen
- Click a credit card button: screen shot 2018-09-05 at 6 04 11 pm
- Verify that <s>name and</s> address fields are prefilled as previously entered

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -464,6 +464,7 @@ class WC_Gateway_PPEC_Client {
 			'order_tax'         => round( WC()->cart->tax_total + WC()->cart->shipping_tax_total, $decimals ),
 			'shipping'          => round( WC()->cart->shipping_total, $decimals ),
 			'items'             => $this->_get_paypal_line_items_from_cart(),
+			'shipping_address'  => $this->_get_address_from_customer(),
 		);
 
 		return $this->get_details( $details, $discounts, $rounded_total, WC()->cart->total );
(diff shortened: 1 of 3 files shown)


jeffstieler:
Is there a trick to getting the first/last name to "stick" with the checkout form? I can't get the name to make it to the PayPal form, and it also doesn't persist when I refresh the checkout page. 🤷‍♂️

dechov:
Good catch – I thought I'd tested those, but seemingly not well enough.

Looks like we'll have to send those values with the payment initialization call (see https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/475), which in all likelihood will also be necessary for https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/477, and save them as customer details.

dechov:
> Looks like we'll have to send those values with the payment initialization call 

Let me know if you favor a different strategy, but as I see it this PR doesn't necessarily need to be expanded to include this. The change to customer data should be made in coordination with the https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/475 branch, and is complementary to this change. (For now, I've put the change to save the customer data [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/eece75b5ee788971026fb0b874b604fa209bc879), in a branch off of that one.)

With this in mind, I've just removed the bit about verifying the prefilled name from the testing instructions above, and acknowledged that it doesn't entirely close https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/474.

includes/class-wc-gateway-ppec-client.php
```diff
@@ -698,6 +699,59 @@ protected function _get_details_from_order( $order_id ) {
...
+	 * @return array Address
+	 */
+	protected function _get_address_from_customer() {
+		$customer = WC()->customer;
+
+		$shipping_address = new PayPal_Address;
+
+		$old_wc = version_compare( WC_VERSION, '3.0', '<' );
+
+		if ( $customer->get_shipping_address() || $customer->get_shipping_address_2() ) {
```
jeffstieler: Do we need to check the 2nd line in the shipping address here? Is there a scenario where a valid address only has something on the 2nd line?

includes/class-wc-gateway-ppec-client.php
```diff
@@ -698,6 +699,59 @@ protected function _get_details_from_order( $order_id ) {
...
+			$shipping_first_name = $old_wc ? '' : $customer->get_shipping_first_name();
+			$shipping_last_name  = $old_wc ? '' : $customer->get_shipping_last_name();
```
jeffstieler: Is there no way to retrieve a customer's name in WC 2.6.x? Can we ask Proton?

includes/class-wc-gateway-ppec-client.php
```diff
@@ -698,6 +699,59 @@ protected function _get_details_from_order( $order_id ) {
...
+		$shipping_address->setStreet2( $shipping_address_2 );
+		$shipping_address->setCity( $shipping_city );
+		$shipping_address->setState( $shipping_state );
+		$shipping_address->setZip( $shipping_postcode );
+
+		// In case merchant only expects domestic shipping and hides shipping
+		// country, fallback to base country.
+		//
+		// @see https://github.com/woothemes/woocommerce-gateway-paypal-express-checkout/issues/139
+		if ( empty( $shipping_country ) ) {
```
jeffstieler: Can we more explicitly check for the scenario where the merchant is only shipping domestically? At first glance it seems like we might be reading too much into an empty shipping country.. would we hit this case if the form wasn't completely filled out?

In the WP.org thread linked in the issue ([#](https://wordpress.org/support/topic/an-error-occured-while-processing-your-paypal-payment?replies=6)), the "sell to" countries setting is referenced. Perhaps we can check that and the "ship to" countries settings to determine the proper action?

includes/class-wc-gateway-ppec-client.php
jeffstieler: 👍 Fair enough

includes/class-wc-gateway-ppec-client.php
jeffstieler: I'm satisfied with that 👍

PR #477, opened 2018-09-06
Pre-fill contact information in PayPal Guest Checkout

dechov:
The "Contact Information" section of the PayPal checkout flow (when paying by credit card) requires both phone and email. These fields are currently not being pre-filled (with or without Smart Payment Buttons enabled), even though we ask this information on the Checkout screen:

Store | PayPal
-- | --
screen shot 2018-09-05 at 10 35 07 pm | screen shot 2018-09-05 at 10 28 24 pm

The changes so far will work only with Smart Payment Buttons disabled.

TODO once https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/476 is merged:
- [ ] make email pre-filling work with Smart Payment Buttons
- [ ] pass phone number to pre-fill, with Smart Payment Buttons

PR #482, opened 2018-09-20
Store checkout form data in customer session, for pre-filling PayPal form

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/474 – the "name" part, which was not fixed by https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/476 because the customer name is not updated in the session when the address changes.

This PR shares a commit (https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/482/commits/b82d9c95e9ff9c9c98a3a0c6180fe38a7a486503) with https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/475, which passes checkout form data in the AJAX request. Here, it is used to store the complete customer data in the session.

(Also a prerequisite for https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/477.)

To test:
- With "Credit Card" not a hidden funding method in the settings
- And with the cart not empty
- Enter values into the billing address fields on the Checkout screen
- Click a credit card button: screen shot 2018-09-05 at 6 04 11 pm
- Verify that first and last name fields are pre-filled on the PayPal side as entered on the store

A little more context on how the form is pre-filled: if available, it is actually the shipping address which is used to pre-fill the form, rather than the billing address, because only one address can be sent to PayPal, and it is as "SHIPTO" fields. It is actually used to pre-fill the shipping address (when starting on screens other than Checkout), and pre-filling the billing address seems to be a secondary consequence.

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -161,6 +161,7 @@ class WC_Gateway_PPEC_Cart_Handler {
 		}
 
 		if ( empty( $error_messages ) ) {
+			$this->set_customer_data( $_POST );
 			$this->start_checkout();
 		} else {
 			wp_send_json_error( array( 'messages' => $error_messages ) );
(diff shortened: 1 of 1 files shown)


includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -182,6 +183,45 @@ protected function start_checkout() {
...
+	 * Store checkout form data in customer session.
+	 *
+	 * @since 1.6.4
+	 */
+	protected function set_customer_data( $data ) {
+		$customer = WC()->customer;
+
+		$shipping_prefix = isset( $data['ship_to_different_address'] ) ? 'shipping' : 'billing';
+
+		$customer->set_shipping_address( $data[ $shipping_prefix . '_address_1' ] );
```
jeffstieler: Let's avoid PHP Warnings by checking for these keys before accessing, and also sanitize the user input.

Using core WC as an example ([#](https://github.com/woocommerce/woocommerce/blob/fd722da88ab5c42afe55965773229e33aec74ca8/includes/class-wc-form-handler.php#L206)):

```php
$shipping_address = empty( $data[ $shipping_prefix . '_address_1' ] ) ? '' : wc_clean( $data[ $shipping_prefix . '_address_1' ] );
$customer->set_shipping_address( $shipping_address );
```

dechov: 👍 Done in f52e141

Issue #486, opened 2018-10-10
Shipping costs are not sent over to PayPal 


fhaps:
On the checkout page, shipping costs are calculated correctly. But when you proceed to make the payment in PayPal, the total there does not include shipping costs.

It includes Product Price + Tax, but it doesn't add the shipping costs.

![](http://cld.wthms.co/KjfeOR+)
Image Link: http://cld.wthms.co/KjfeOR

Some numbers above might be hard to see, so here's a clearer screenshot of the shipping costs plus order total:

![](http://cld.wthms.co/aJHVi9+)
Image Link: http://cld.wthms.co/aJHVi9

Tested this on both AT and self-hosted.

Affecting 1484712-zed.

mcaiazzo23:
Having the same issue. Temporary fix was commenting this filter out

`add_filter( 'woocommerce_cart_needs_shipping', '__return_false' );`

on line 129 of /includes/class-wc-gateway-ppec-cart-handler.php

fhaps:
**New findings:** The issue appears when the `Smart Payment Buttons` are enabled. If these are disabled, the issue is gone and shipping costs are correctly carried over to PayPal.

With Smart Payment Buttons _enabled_:

![](http://cld.wthms.co/tKJDCR+)
Image Link: http://cld.wthms.co/tKJDCR

With Smart Payment Buttons _disabled_:

![](http://cld.wthms.co/y40erJ+)
Image Link: http://cld.wthms.co/y40erJ

zwiesman31:
Here's another 1501259-zen

CC: @dechov

yukikatayama:
1502710-zen another report

luistcode:
waiting a solution for this. Thanks.

MrNghia123:
Same problem.

bnovina:
The issue appears when the Smart Payment Buttons are enabled only. Anyway, the shipping costs will appear in Activity of PayPal account.
However, there is additional issue when the Smart Payment Buttons are enabled. PayPal does not recognize the country specified in the checkout billing form and return completely different country.
No shipping costs, and wrong country. UK instead of Germany.
![example](https://user-images.githubusercontent.com/44400200/47359377-e3e66600-d6cc-11e8-93ef-0588fe133be9.png)

AashikP:
[...]

PR #489, opened 2018-10-23
Properly handle shipping address flag and fix missing shipping total

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/486

As pointed out in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/486#issuecomment-429113321, removing the filter on `woocommerce_cart_needs_shipping` solves the issue of the missing shipping total (implemented in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/5c3e09856e59d57ffbd639301ddfe941657fee8d).

The filter was originally added so as to not show a shipping address selector on the PayPal checkout screen when starting on the store checkout screen (and having already entered a shipping address). Besides being overly broad, this was not the best way to handle the shipping address, as it affects the `NOSHIPPING` flag rather than the `ADDROVERRIDE` flag (see [API docs](https://developer.paypal.com/docs/classic/api/merchant/SetExpressCheckout_API_Operation_NVP/)). The latter is intended for this case, showing the shipping address that was passed rather than no shipping address at all.

Tested by confirming (with Smart Payment Buttons both enabled and disabled) that the cart page allows address selection, while the checkout page just displays the previously entered address, and verifying that the shipping cost is displayed.

The only change should be when using Smart Payment Buttons from the checkout screen:
screen shot 2018-10-23 at 6 58 06 pm

When clicking "Change":
screen shot 2018-10-23 at 6 58 26 pm

When clicking on price total, the "Shipping" total is present:
screen shot 2018-10-23 at 6 58 54 pm

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -227,8 +227,8 @@ class WC_Gateway_PPEC_Client {
 	 * @param array $args {
 	 *     Context args to retrieve SetExpressCheckout parameters.
 	 *
-	 *     @type string $start_from               Start from 'cart' or 'checkout'.
-	 *     @type int    $order_id                 Order ID if $start_from is 'checkout'.
+	 *     @type string $skip_checkout            Whether checking out ahead of store checkout screen.
+	 *     @type int    $order_id                 Order ID if checking out after order is created.
 	 *     @type bool   $create_billing_agreement Whether billing agreement creation
 	 *                                            is needed after returned from PayPal.
 	 * }
@@ -239,7 +239,7 @@ class WC_Gateway_PPEC_Client {
 		$args = wp_parse_args(
 			$args,
 			array(
-				'start_from'               => 'cart',
+				'skip_checkout'            => true,
 				'order_id'                 => '',
 				'create_billing_agreement' => false,
 			)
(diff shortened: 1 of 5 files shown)


includes/class-wc-gateway-ppec-client.php
```diff
@@ -261,7 +261,7 @@ public function get_set_express_checkout_params( array $args ) {
 			$params['USERSELECTEDFUNDINGSOURCE'] = 'Finance';
 		}
 
-		if ( 'checkout' === $args['start_from'] ) {
+		if ( ! $args['skip_checkout'] ) {
```
DanReyLop: Nitpick: A comment here explaining what does `ADDROVERRIDE` mean would be quite useful (or just a comment in the function header pointing to the API docs)

dechov:
> you can do a regular old PayPal Checkout in the cart page without actually calculating the shipping cost, and in that case it's $0. Is that a regression?

As far as I can tell, this is true regardless of Smart Payment Buttons being enabled, and (as I've just confirmed) pre-1.6 behaves the same as well. I don't think it's a problem, as the eventual payment goes through with the correct shipping amount shown on the checkout page – which is also the case in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/486.

DanReyLop:
> As far as I can tell, this is true regardless of Smart Payment Buttons being enabled, and (as I've just confirmed) pre-1.6 behaves the same as well.

Cool, thanks for the clarification! :shipit:

Issue #462, opened 2018-08-01
checkout.js loaded when not needed.


geeklore:
Even with option unchecked `**Checkout on Single Product** - `https://www.paypalobjects.com/api/checkout.js` is being loaded on every page of the site.

Customer reported this because their users are all on a G3 network and every file loaded counts.

Can we make it so that it's conditionally loaded?

I feel this is more of a question, but making it a bug so it gets some attention.

geeklore:
zen-1319182

SiR-DanieL:
hi @bor0 any idea when you'll be able to look at this? I see this is focus extension this week

bor0:
You need to disable `Smart Payment Buttons` if you want to not load this JS file, as SPB needs it.

juliaamosova:
Pinging @woocommerce/extendables here so this issue can be looked at and triaged.

lkraav:
> You need to disable Smart Payment Buttons if you want to not load this JS file, as SPB needs it (but it looks like it will enqueue anyway for the cart page).

Is it not possible to detect whether there's any SPB-s actually on the page, and only enqueue then?

treibalen:
@woocommerce/extendables pinging as it seems that the previous reply needs dev's attention :)

jessepearson:
@woocommerce/hydra Since this is now yours. It appears that the js files are loaded regardless if they are needed or not, is there any way around this?

DanReyLop:
@jessepearson We've just released version 1.6.5 of the extension, it includes a fix for this issue.

PR #491, opened 2018-10-25
Only enqueue the checkout and payment buttons scripts when needed

DanReyLop:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/462

I've changed some `wp_enqueue_script()` to `wp_register_script()`, and moved the actual enqueueing of the scripts to the moment the Smart Paypal Button containers are rendered. The result is that the Smart PayPal Buttons will keep working normally, and no extra JS will be included if the buttons aren't rendered.

@dechov I've tried using Storefront, the buttons load correctly in the product page, cart, and checkout. Is there anything that could go wrong? What kind of caching issues did you encounter?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ deploy-to-wp-org.sh
 node_modules
 Gruntfile.js
 package.json
+.idea
+*.iml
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@ deploy-to-wp-org.sh
 node_modules
 Gruntfile.js
 package.json
+.idea
+*.iml
```
dechov: Were these added intentionally?

DanReyLop: Yep, those are used by WebStorm. We have the same ignores in `woocommerce-services`, it's unrelated with this change but it doesn't hurt :)

PR #494, opened 2018-11-02
Update gateway settings for Canadian english locale

arcticlinux:
Added en_CA locale in order to get the paypal express checkout to use Canada as the default checkout country when wordpress locale is set to en_CA for Canada

### DIFF
--- a/includes/class-wc-gateway-ppec-settings.php
+++ b/includes/class-wc-gateway-ppec-settings.php
@@ -25,6 +25,7 @@ class WC_Gateway_PPEC_Settings {
 		'da_DK',
 		'de_DE',
 		'en_AU',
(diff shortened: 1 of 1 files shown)


arcticlinux:
This requires changing the Wordpress language to en_CA as well.  Settings -> General -> Site Language -> English (Canada)

This doesn't appear to match the Paypal documentation https://developer.paypal.com/docs/classic/api/locale_codes/#supported-locale-codes which lists Canada english as en_US, but in testing this did work properly, I suspect the documentation is wrong, not sure how to bring that to their attention.

menakas:
Closing this in favour of #536 which includes 'en_CA' as well. The progress can be tracked there.

PR #503, opened 2018-11-19
Update WP compatibility

bor0:
Friendly reminder to update this. Needs to be released with WP 5.0.

cc @dechov


bor0:
Closing stale PR. It still can be used as a reference if needed.

PR #507, opened 2018-11-25
Fix using the PayPal smart button on a variable product page, or a product with Addons

### LINKED ISSUE #467: With require Product Addons, can checkout from product page without selecting the required addons

### LINKED ISSUE #500: Unable to checkout a variable product using Smart Button

DanReyLop:
Fixes #500 
Fixes #467

I've merged both fixes into the same PR because they are very related and the amount of code is still small enough for a single PR.

### Fix for #500 (Unable to buy variation from product page)

This PR includes a more complete version of the fix on https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/506.

When making any kind of request to the server, WooCommerce runs this code: https://github.com/woocommerce/woocommerce/blob/897af8d20ef24e58164376ef65d980d2759f5956/includes/class-wc-form-handler.php#L706

If there's a GET/POST variable called `add-to-cart`, it will try to add that product to the cart. The request structure is different from what WC Core expects, so for variable products it won't find the required attributes and it will throw [an exception](https://github.com/woocommerce/woocommerce/blob/3.5.1/includes/class-wc-form-handler.php#L869).

For the most part it doesn't matter because the cart contents are cleaned up anyway before PayPal's `wc_ajax_generate_cart` will add the products to the cart correctly. But the exception thrown is persisted, so after coming back from a successful PayPal flow, [this logic](https://github.com/woocommerce/woocommerce/blob/ca08beaa42d396681cfa69c5222807702a77abf1/includes/shortcodes/class-wc-shortcode-checkout.php#L262) will kick in and display an error page.

To fix it, I simply renamed the `add-to-cart` POST parameter to `ppec-add-to-cart` so it doesn't clash with that built-in WooCommerce functionality. Should help with performance a bit too.

#### To test:
* Enable SPB in product pages.
* Go to a variable product page.
* Pick a variation.
* Click the PayPal button.
* Go through the whole PayPal flow (you don't need to actually pay, so no need to use a sandbox account).
* When you're redirected back to the "Checkout - Order review" page, you shouldn't see any errors and you should see the button to confirm the order.### Fix for #500 (Unable to buy variation from product page)

### Fix for #467 (can use PayPal from product page without inputting required fields)

Product Addons adds plain HTML5 form fields to the product page. Those fields can have a `required` flag or even more complex validation. The `Add to cart` is a plain `<input type=submit>` button, so in modern browsers ([anything newer than IE10](https://caniuse.com/#feat=constraint-validation)), that validation is run automatically before attempting to submit the form. I've added equivalent logic on JS, checking the validity of every field on the form and disabling the PayPal button if there's any invalid one.

This PR also includes better integration (i.e. not broken) with Product Addons. Product Addons adds some merchant-configurable fields to the product page, and those fields are sent to the server when the item is added to the cart.

The solution here is to preserve the same logic while sending the "generate cart" AJAX request that happens when clicking the PayPal smart button.

#### To test:
* Enable SPB in product pages.
* Install Product Addons, add some optional and/or requried fields to the product you're testing this with.
* Go to the product page.
* Check that, unless you fill all the required fields, the PayPal button will be disabled and unresponsive.
* Go through the whole PayPal flow (you don't need to actually pay, so no need to use a sandbox account).
* When you're redirected back to the "Checkout - Order review" page, you shouldn't see any errors, you should see the button to confirm the order, and you should see on the order summary all the values for the fields you entered.

This solution is not 100% ideal because the behaviour of the PayPal button is not 100% consistent with the `Add to cart` button (the `Add to Cart` button stays enabled but the form will display errors if you click it). But for now I think it's a net improvement.

### DIFF
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -86,10 +86,8 @@
 			data[ field_pairs[ i ].name ] = field_pairs[ i ].value;
 		}
 
-		if ( ! data[ 'ppec-add-to-cart' ] ) {
-			// If this is a simple product, the "Submit" button has the product ID as "value", we need to include it explicitly
-			data[ 'ppec-add-to-cart' ] = $( '[name=add-to-cart]' ).val();
-		}
+		// If this is a simple product, the "Submit" button has the product ID as "value", we need to include it explicitly
+		data[ 'ppec-add-to-cart' ] = $( '[name=add-to-cart]' ).val();
 
 		$.ajax( {
 			type:    'POST',
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -30,20 +30,34 @@
 			$( '#woo_pp_ec_button_product > *' ).css( 'pointer-events', 'none' );
 		} );
 
+    var variation_valid = true;
+    var fields_valid = true;
+    var update_button = function() {
+        $( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
+	};
+
+    var validate_form = function() {
+        silent_validation = true;
+        fields_valid = true;
+        form.find( 'select:enabled, input:enabled, textarea:enabled' ).each( function() {
+            fields_valid = fields_valid && this.checkValidity();
+        } );
+        silent_validation = false;
+        update_button();
+    };
+
 	// It's a variations form, button availability should depend on its events
-	var variation_valid = true;
 	if ( $( '.variations_form' ).length ) {
-		$( '#woo_pp_ec_button_product' ).trigger( 'disable' );
 		variation_valid = false;
 
 		$( '.variations_form' )
 		.on( 'show_variation', function( event, form, purchasable ) {
-			$( '#woo_pp_ec_button_product' ).trigger( purchasable ? 'enable' : 'disable' );
 			variation_valid = purchasable;
+            update_button();
 		} )
 		.on( 'hide_variation', function() {
-			$( '#woo_pp_ec_button_product' ).trigger( 'disable' );
 			variation_valid = false;
+            update_button();
 		} );
 	}
 
@@ -55,18 +69,8 @@
 			e.preventDefault();
 		}
 	}, true );
-	form.on( 'change', 'select, input, textarea', function() {
-		if ( ! variation_valid ) {
-			return;
-		}
-		silent_validation = true;
-		var valid = true;
-		form.find( 'select:enabled, input:enabled, textarea:enabled' ).each( function() {
-			valid = valid && this.checkValidity();
-		} );
-		$( '#woo_pp_ec_button_product' ).trigger( valid ? 'enable' : 'disable' );
-		silent_validation = false;
-	} );
+	form.on( 'change', 'select, input, textarea', validate_form );
+	validate_form();
 
 	var get_attributes = function() {
 		var select = $( '.variations_form' ).find( '.variations select' ),
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -38,10 +38,7 @@
 
 	var validate_form = function() {
 		silent_validation = true;
-		fields_valid = true;
-		form.find( 'select:enabled, input:enabled, textarea:enabled' ).each( function() {
-			fields_valid = fields_valid && this.checkValidity();
-		} );
+		fields_valid = form.get( 0 ).checkValidity();
 		silent_validation = false;
 		update_button();
 	};
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -30,21 +30,21 @@
 			$( '#woo_pp_ec_button_product > *' ).css( 'pointer-events', 'none' );
 		} );
 
-    var variation_valid = true;
-    var fields_valid = true;
-    var update_button = function() {
-        $( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
+	var variation_valid = true;
+	var fields_valid = true;
+	var update_button = function() {
+		$( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
 	};
 
-    var validate_form = function() {
-        silent_validation = true;
-        fields_valid = true;
-        form.find( 'select:enabled, input:enabled, textarea:enabled' ).each( function() {
-            fields_valid = fields_valid && this.checkValidity();
-        } );
-        silent_validation = false;
-        update_button();
-    };
+	var validate_form = function() {
+		silent_validation = true;
+		fields_valid = true;
+		form.find( 'select:enabled, input:enabled, textarea:enabled' ).each( function() {
+			fields_valid = fields_valid && this.checkValidity();
+		} );
+		silent_validation = false;
+		update_button();
+	};
 
 	// It's a variations form, button availability should depend on its events
 	if ( $( '.variations_form' ).length ) {
@@ -53,11 +53,11 @@
 		$( '.variations_form' )
 		.on( 'show_variation', function( event, form, purchasable ) {
 			variation_valid = purchasable;
-            update_button();
+			update_button();
 		} )
 		.on( 'hide_variation', function() {
 			variation_valid = false;
-            update_button();
+			update_button();
 		} );
 	}
 
@@ -77,14 +77,14 @@
 			'nonce': wc_ppec_generate_cart_context.generate_cart_nonce,
 		};
 
-        var field_pairs = form.serializeArray();
-        for ( var i = 0; i < field_pairs.length; i++ ) {
-        	// Prevent the default WooCommerce PHP form handler from recognizing this as an "add to cart" call
-            if ( 'add-to-cart' === field_pairs[ i ].name ) {
-                field_pairs[ i ].name = 'ppec-add-to-cart';
+		var field_pairs = form.serializeArray();
+		for ( var i = 0; i < field_pairs.length; i++ ) {
+			// Prevent the default WooCommerce PHP form handler from recognizing this as an "add to cart" call
+			if ( 'add-to-cart' === field_pairs[ i ].name ) {
+				field_pairs[ i ].name = 'ppec-add-to-cart';
 			}
 			data[ field_pairs[ i ].name ] = field_pairs[ i ].value;
-        }
+		}
 
 		$.ajax( {
 			type:    'POST',
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -33,8 +33,10 @@
 	// True if the product is simple or the user selected a valid variation. False on variable product without a valid variation selected
 	var variation_valid = true;
 
-	// True if all the fields of the product form are valid. False otherwise
+	// True if all the fields of the product form are valid (such as required fields configured by Product Add-Ons). False otherwise
 	var fields_valid = true;
+
+	var form = $( 'form.cart' );
 
 	var update_button = function() {
 		$( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
@@ -61,7 +63,6 @@
 	}
 
 	// Disable the button if there are invalid fields in the product page (like required fields from Product Addons)
-	var form = $( 'form.cart' );
 	form.on( 'change', 'select, input, textarea', function() {
 		// Hack: IE11 uses the previous field value for the checkValidity() check if it's called in the onChange handler
 		setTimeout( validate_form, 0 );
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -77,9 +77,9 @@
 			'nonce': wc_ppec_generate_cart_context.generate_cart_nonce,
 		};
 
-        var field_pairs = $( form ).serializeArray();
+        var field_pairs = form.serializeArray();
         for ( var i = 0; i < field_pairs.length; i++ ) {
-        	// Prevent the default WooCommerce PHP form handler to recognize this as an "add to cart" call
+        	// Prevent the default WooCommerce PHP form handler from recognizing this as an "add to cart" call
             if ( 'add-to-cart' === field_pairs[ i ].name ) {
                 field_pairs[ i ].name = 'ppec-add-to-cart';
 			}
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -31,17 +31,42 @@
 		} );
 
 	// It's a variations form, button availability should depend on its events
+	var variation_valid = true;
 	if ( $( '.variations_form' ).length ) {
 		$( '#woo_pp_ec_button_product' ).trigger( 'disable' );
+		variation_valid = false;
 
 		$( '.variations_form' )
 		.on( 'show_variation', function( event, form, purchasable ) {
 			$( '#woo_pp_ec_button_product' ).trigger( purchasable ? 'enable' : 'disable' );
+			variation_valid = purchasable;
 		} )
 		.on( 'hide_variation', function() {
 			$( '#woo_pp_ec_button_product' ).trigger( 'disable' );
+			variation_valid = false;
 		} );
 	}
+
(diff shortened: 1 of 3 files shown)


v18:
On my local test site I was able to see this error in my console from PayPal in both cases: `serialize.js:175 Uncaught Error: Error: No value passed to payment` as well as `logger.js:63 ppxo_no_token_passed_to_payment`

Didn't have a chance to dig much further yet.

DanReyLop:
@marcinbot I finally got around to trying this. Great catch! Turns out that for simple products, the product ID is sent from the `<button type="submit">` instead of an `<input type="hidden">`. I haven't even considered it because, frankly, it makes no sense. Your change works and it's a 👍 from me.

@v18 Can you still reproduce the "Variable products" case? If not, I think we can safely go ahead with this PR and release a new version after the Christmas freeze.

v18:
> This solution is not 100% ideal because the behaviour of the PayPal button is not 100% consistent with the Add to cart button (the Add to Cart button stays enabled but the form will display errors if you click it). But for now I think it's a net improvement.

Agree here. I opened an issue:

https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/515

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,67 @@
...
+	var fields_valid = true;
+	var update_button = function() {
+		$( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
+	};
+
+	var validate_form = function() {
+		silent_validation = true;
+		fields_valid = true;
+		form.find( 'select:enabled, input:enabled, textarea:enabled' ).each( function() {
+			fields_valid = fields_valid && this.checkValidity();
```
v18: Is there a reason not use `checkValidity()` on the entire form instead of on each of the input elements?

For example, `form.get( 0 ).checkValidity()`?

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,67 @@
...
-		return {
-			'count'      : count,
-			'chosenCount': chosen,
-			'data'       : data
-		};
-	};
+	// Disable the button if there are invalid fields in the product page (like required fields from Product Addons)
+	var silent_validation;
+	var form = $( 'form.cart' );
+	form.get( 0 ).addEventListener( 'invalid', function( e ) {
```
v18: I'm not sure I understand why this event listener is added. Can you explain?

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,67 @@
...
+		if ( silent_validation ) {
+			e.preventDefault();
+		}
+	}, true );
+	form.on( 'change', 'select, input, textarea', validate_form );
```
v18: Did you include `select` and `textarea` to catch all non-`<input>` elements of a form, so we cover all other cases other than the current version of Product Addons?

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,67 @@
...
+		var field_pairs = form.serializeArray();
+		for ( var i = 0; i < field_pairs.length; i++ ) {
+			// Prevent the default WooCommerce PHP form handler from recognizing this as an "add to cart" call
+			if ( 'add-to-cart' === field_pairs[ i ].name ) {
+				field_pairs[ i ].name = 'ppec-add-to-cart';
+			}
+			data[ field_pairs[ i ].name ] = field_pairs[ i ].value;
+		}
+
+		if ( ! data[ 'ppec-add-to-cart' ] ) {
```
v18: Why are we checking for the existence of `data[ 'ppec-add-to-cart' ]` first?

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,67 @@
...
-		} );
-
-		return {
-			'count'      : count,
-			'chosenCount': chosen,
-			'data'       : data
-		};
-	};
+	// Disable the button if there are invalid fields in the product page (like required fields from Product Addons)
+	var silent_validation;
```
v18: Can we move these types of variable initializations and declarations closer to the top with some explanation comment for each, so that they're not lost or accidentally removed later? E.g. `silent_validation`, `variation_valid`, `fields_valid`.

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,67 @@
...
 	var generate_cart = function( callback ) {
 		var data = {
-			'nonce':       wc_ppec_generate_cart_context.generate_cart_nonce,
-			'qty':         $( '.quantity .qty' ).val(),
-			'attributes':  $( '.variations_form' ).length ? get_attributes().data : [],
-			'add-to-cart': $( '[name=add-to-cart]' ).val(),
+			'nonce': wc_ppec_generate_cart_context.generate_cart_nonce,
 		};
 
+		var field_pairs = form.serializeArray();
```
v18: I'm not sure I understand why we replace `get_attributes()` with the loop below. With `get_attributes` we cared only about `.variations_form` and not the entire form. 

Can you explain what `data` we expected before, what we now expect so see in `data`, and why the changes are needed?

v18:
@marcinbot @DanReyLop the changes in this PR now work for me a well 👍. I reviewed the code now as well.

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: The reason is that I thought `checkValidity` is only for individual fields (that's what's referred to in almost all the documentation). There's a similar [`form.get(0).reportValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reportValidity) method, but [it's not supported in IE 11](https://caniuse.com/#feat=constraint-validation).

I've tried `form.get( 0 ).checkValidity()` on https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/507/commits/347195ab13863fbc9a536c2b60f433c0b1363e99 and it works. Great catch!

Also, while testing this, I fired up the old IE11 VM and turns out the validation didn't work as it should. Fixed in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/507/commits/3f76659905474f0fb07d8df031baa9bdc5cba8cf

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: This listener blocks the `oninvalid` event from reaching "the browser". Without it, an annoying "This field is required" pop-over would appear over the invalid fields every time another field changes. That's why I called the flag `silent_validation`.

DanReyLop: Adding to that, an `oninvalid` event is fired on a field when:
- The user clicks the "Submit" button and the field fails validation.
- `inputElement.checkValidity()` is called, and the field fails validation.

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: As you pointed out, `get_attributes()` only cared about the variations fields. By serializing the whole form, you get those variation fields **and** any fields added by a plugin **and** the basic fields like `quantity` and `add-to-cart` (product ID).

This logic is simpler, and it basically sends on a POST request the same data that would be sent in a plain HTML form (if JS was disabled). Product Add-Ons relies on that assumption, for example (it fetches the values directly from `$_POST` whenever `WC()->cart->add_to_cart()` is called).

That's why I merged the 2 issues into this 1 PR, because the fi for both variations misbehaving and Product Add-ons compatibility is basically the same: send all the HTML form data to the `wc_ajax_generate_cart()` AJAX endpoint.

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: Hum... I suppose that's redundant, there's no harm in always overwriting `data[ 'ppec-add-to-cart' ]` even if the product is a variable product. Removed in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/507/commits/dc8c28341409226ea4a069b794eaa159ed10c26c

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: Product Addons already supports `<textarea>` and `<select>`, so this wasn't just me thinking about "the future" :)

assets/js/wc-gateway-ppec-generate-cart.js
v18: Ah, I see what you mean - since we're using `checkValidity` and it can fire an `invalid` event, we should listen to it and cancel the HTML 5 validation that can be displayed.

I did a bit of digging, and from what I can see the `invalid` event is not what prompts the browser to display the HTML 5 validation errors. I also tested this by removing the `e.preventDefault` line, and I see that the validation errors are not displayed when I change any of the form fields. This works on all browsers (didn't personally check IE11).

So, it looks like `e.preventDefault` is no needed for `invalid` events. The form will display errors only when it's submitted.

Is there another issue I'm missing?

v18:
As part of my digging into form validation ([#](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/507#discussion_r242575505)), I also found out that most^ browsers now support `reportValidity()`, which can be used to display the HTML5 errors that submitting a form displays. We can use it to solve #515 and mimic the behaviour of the submit button. I guess it doesn't have to be done in this PR, but since it's easier than expected can we do a quick follow-up PR?

^ - this applies to all the browsers we can about except IE11. For IE11 we can use our current strategy of disabling the PayPal buttons completely, and add an error message. Graceful degradation ftw!

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: Hah! Good catch. My first try was to use `reportValidity()`, and that **did** prompt the HTML5 validation pop-overs. That's not needed with `checkValidity()`. I ❤️ these APIs that are almost the same but different in sutble ways...

I've removed the whole listener in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/507/commits/b61ec7d1fa7d09c346fd5391f400c4280002992e

v18:
> but the PayPal flow would open regardless. Once you click the PayPal button, nothing can stop the dialog to pop-up.

Ooh, I see! Thanks for the details. Definitely let's punt to another PR then 👍

assets/js/wc-gateway-ppec-generate-cart.js
v18: Thanks for explaining! I got it now 😄

assets/js/wc-gateway-ppec-generate-cart.js
v18: I know that feeling... :)

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,66 @@
...
 		} );
 
+	// True if the product is simple or the user selected a valid variation. False on variable product without a valid variation selected
+	var variation_valid = true;
+
+	// True if all the fields of the product form are valid. False otherwise
+	var fields_valid = true;
+
+	// True if there's an automatic form validation being performed (not user-initiated), so the user must not be alerted. False otherwise
+	var silent_validation = false;
```
v18: Since we're using `checkValidity()` and it doesn't show show validation errors to the user, we can remove `silent_validation` from all the parts of the code, right?

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,66 @@
...
-		} );
-
-		return {
-			'count'      : count,
-			'chosenCount': chosen,
-			'data'       : data
-		};
-	};
+	// Disable the button if there are invalid fields in the product page (like required fields from Product Addons)
+	var form = $( 'form.cart' );
```
v18: What do you think of moving the `form` definition closer to the top as well? It's being used in `validate_form()` above so it would be greater for the reader to see it earlier.

assets/js/wc-gateway-ppec-generate-cart.js
v18: Same for lines 47 & 49 below.

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,66 @@
 			$( '#woo_pp_ec_button_product > *' ).css( 'pointer-events', 'none' );
 		} );
 
+	// True if the product is simple or the user selected a valid variation. False on variable product without a valid variation selected
+	var variation_valid = true;
+
+	// True if all the fields of the product form are valid. False otherwise
```
v18: Probably useful to add to this comment that this came as support for Product Addons, which adds form fields to the cart.

assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -30,52 +30,66 @@
...
-			'count'      : count,
-			'chosenCount': chosen,
-			'data'       : data
-		};
-	};
+	// Disable the button if there are invalid fields in the product page (like required fields from Product Addons)
+	var form = $( 'form.cart' );
+	form.on( 'change', 'select, input, textarea', function() {
+		// Hack: IE11 uses the previous field value for the checkValidity() check if it's called in the onChange handler
+		setTimeout( validate_form, 0 );
```
v18: Wow, great job finding that 🤩I also noticed that grayscale CSS doesn't work for IE11, that's too bad. There isn't an easy way to get it to work, so the 🚫 icon folks see now on hover will have to do for now.

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: 🤦‍♂️ Yes. Done

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: Yeah, I give up on IE11 being pizel-perfect, this will have to do :)

assets/js/wc-gateway-ppec-generate-cart.js
DanReyLop: Yeah, done 👍

v18:
Looking good @DanReyLop ! Ready to :shipit: . Before merging, could you rebase so the commit history is a bit cleaner and easier to read as a good outline of the changes that took place?

DanReyLop:
> Before merging, could you rebase so the commit history is a bit cleaner and easier to read as a good outline of the changes that took place?

It's not possible because at some point I merged with `master` because of conflicts, instead of rebasing (I did it from the GitHub UI). I'll squash and edit the commit message so it makes sense :)

v18:
Hey hey, no worries. The squash commit is definitely better than the 19 commits we had 😌 😄 .

By the way, it's still generally possible to rebase even with a merge commit. I've run into some hairy conflicts where I gave up, though, for sure.

In this particular case, there was only a small conflict in the readme file so the rebase would have worked. Here's the flow I used to see how it might play out:

`git checkout update/support-product-addons`
`git pull --rebase origin c7476d` (this is the last commit on master before merging this PR)

There was a merge conflict in readme.txt, which I fixed manually and staged the changes.

`git rebase --continue`

That made all the commits on this branch sit on top of `c7476d`. From here, the commits can be changed using interactive `git rebase`.

DanReyLop:
> By the way, it's still generally possible to rebase even with a merge commit.

`master` was merged into this branch... and now you can rebase it *onto* `master`... I just assumed it wasn't possible and didn't try it, it requires some 4th-dimensional thinking right there 😂 

> The squash commit is definitely better than the 19 commits we had

Yeah, in this case I would have squashed anyway. Most of these changes were quite small and back-tracking previous commits. I don't want all that noise in `master` :)

PR #509, opened 2018-11-30
Add process_refund void logic for authorize only transaction

creativeround:
Fix allows for the admin to refund a non-captured authorize only transaction by doing a void instead of refund in the process_refund.

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -367,6 +367,26 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 		$txn_data = $old_wc ? get_post_meta( $order_id, '_woo_pp_txnData', true ) : $order->get_meta( '_woo_pp_txnData', true );
 		$order_currency = $old_wc ? $order->order_currency : $order->get_currency();
 
+		
+		// Process a void instead of refund if transaction is authorized only	
+		$trans_id = get_post_meta( $order_id, '_transaction_id', true );
+		$trans_details = wc_gateway_ppec()->client->get_transaction_details( array( 'TRANSACTIONID' => $trans_id ) );
+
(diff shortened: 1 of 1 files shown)


allendav:
@dechov would you take a look at this? - voiding auths is not an uncommon workflow for merchants Thanks!

dechov:
@creativeround Thanks for this PR! It seems the current way to void the authorization is to cancel the "On hold" order (i.e. set status to "Cancelled") or refund its full amount manually (at which point status is set to "Refunded"). Is it correct to say this PR is to add support for another way, which is via the "Refund […] via PayPal Checkout" button?

I tested by refunding the full amount in this way, and it works well. However, if I put less than the total as the refund amount, it voids the entire transaction even though only a partial refund is reflected on the order screen. What do you think about hiding this input for this "void" case? (The "Reason for refund" can be either hidden or passed as the `NOTE` field in the `DoVoid` request.)

Also, in that case, since we would be making special accommodation for this "On hold" case anyway, can we perhaps change the action from "Refund" to "Void"?

An alternative may be to disable the "Refund […] via PayPal Checkout" button altogether for "On hold" orders, and try to make it more evident that cancelling the order would void the authorization.

includes/abstracts/abstract-wc-gateway-ppec.php
```diff
@@ -367,6 +367,26 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
 		$txn_data = $old_wc ? get_post_meta( $order_id, '_woo_pp_txnData', true ) : $order->get_meta( '_woo_pp_txnData', true );
 		$order_currency = $old_wc ? $order->order_currency : $order->get_currency();
 
+		
+		// Process a void instead of refund if transaction is authorized only
```
dechov: Trailing tab on this line, as well as the blank lines before and after this function.

dechov:
Closing for now, as authorization can be voided with status change, but feel free to re-open if needed.

PR #510, opened 2018-12-06
Display PayPal fees under the totals on the order admin page

### LINKED ISSUE #504: Not Returning PayPal Transaction Fees in Order Data
The Stripe Gateway plugin actually returns back to WooCommerce the Stripe Fees so I can grab it with things like Zapier so I can automate the fee expense into Quickbooks. Looking at the IPN history from orders it appears that the fee is sent back to WooCommerce it is just not logged anywhere on the order like stripe does.

Is this something that could be added if the return is already including it? I think it could be very helpful to others even if so.

Would really help if it showed these like the Stripe Screen Cap. (https://www.screencast.com/t/tSflMXRMs)

I saw these tickets related to Fees but nothing specific to this.  I have confirmed on multiple orders that Zapier sees no fee data in the order meta sent over like it sees with stripe it's just not there.  Thanks in advance!

https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/289
https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/414

marcinbot:
Fixes #504 

screen shot 2018-12-06 at 14 17 03

Test steps:
* if you're using a local site, make sure you're running ngrok
* purchase an item and pay using the PPEC
* log into the merchant dashboard on PayPal and accept the payment - paypal should sand an update to your site containing the fees
* open the order in wp-admin and verify that the fees appear

### DIFF
--- a/includes/class-wc-gateway-ppec-admin-handler.php
+++ b/includes/class-wc-gateway-ppec-admin-handler.php
@@ -311,9 +311,8 @@ class WC_Gateway_PPEC_Admin_Handler {
 		$order = wc_get_order( $order_id );
 
 		$old_wc         = version_compare( WC_VERSION, '3.0', '<' );
-		$order_id       = $old_wc ? $order->id : $order->get_id();
 		$payment_method = $old_wc ? $order->payment_method : $order->get_payment_method();
-		$paypal_fee     = $old_wc ? get_post_meta( $order_id, 'PayPal Transaction Fee', true ) : $order->get_meta( 'PayPal Transaction Fee', true );
+		$paypal_fee     = wc_gateway_ppec_get_transaction_fee( $order );
 		$order_currency = $old_wc ? $order->order_currency : $order->get_currency();
 		$order_total    = $old_wc ? $order->order_total : $order->get_total();
 
--- a/includes/class-wc-gateway-ppec-admin-handler.php
+++ b/includes/class-wc-gateway-ppec-admin-handler.php
@@ -313,7 +313,7 @@ class WC_Gateway_PPEC_Admin_Handler {
 		$old_wc         = version_compare( WC_VERSION, '3.0', '<' );
 		$order_id       = $old_wc ? $order->id : $order->get_id();
 		$payment_method = $old_wc ? $order->payment_method : $order->get_payment_method();
-		$paypal_fee     = get_post_meta( $order_id, 'PayPal Transaction Fee', true );
+		$paypal_fee     = $old_wc ? get_post_meta( $order_id, 'PayPal Transaction Fee', true ) : $order->get_meta( 'PayPal Transaction Fee', true );
 		$order_currency = $old_wc ? $order->order_currency : $order->get_currency();
 		$order_total    = $old_wc ? $order->order_total : $order->get_total();
 
--- a/includes/class-wc-gateway-ppec-admin-handler.php
+++ b/includes/class-wc-gateway-ppec-admin-handler.php
@@ -336,7 +336,7 @@ class WC_Gateway_PPEC_Admin_Handler {
 		</tr>
 		<tr>
 			<td class="label ppec-payout">
-				<?php echo wc_help_tip( __( 'This represents the net total that will be credited to your PayPal account. This may be in the currency that is set in your PayPal account.', 'woocommerce-gateway-paypal-express-checkout' ) ); ?>
+				<?php echo wc_help_tip( __( 'This represents the net total that will be credited to your PayPal account. This may be in a different currency than is set in your PayPal account.', 'woocommerce-gateway-paypal-express-checkout' ) ); ?>
 				<?php esc_html_e( 'PayPal Payout:', 'woocommerce-gateway-paypal-express-checkout' ); ?>
 			</td>
 			<td width="1%"></td>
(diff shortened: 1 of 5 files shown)


includes/class-wc-gateway-ppec-admin-handler.php
```diff
@@ -297,4 +299,53 @@ public function maybe_reset_api_credentials() {
...
+				<?php esc_html_e( 'PayPal Fee:', 'woocommerce-gateway-paypal-express-checkout' ); ?>
+			</td>
+			<td width="1%"></td>
+			<td class="total">
+				-&nbsp;<?php echo wc_price( $paypal_fee, array( 'currency' => $order_currency ) ); ?>
+			</td>
+		</tr>
+		<tr>
+			<td class="label ppec-payout">
+				<?php echo wc_help_tip( __( 'This represents the net total that will be credited to your PayPal account. This may be in the currency that is set in your PayPal account.', 'woocommerce-gateway-paypal-express-checkout' ) ); ?>
```
allendav: Did you mean "may **not** be in the currency"?

includes/class-wc-gateway-ppec-ipn-handler.php
```diff
@@ -166,9 +166,10 @@ protected function validate_amount( $order, $amount ) {
 	 * @param array $posted_data Posted data
 	 */
 	protected function payment_status_completed( $order, $posted_data ) {
-		$order_id = version_compare( WC_VERSION, '3.0', '<' ) ? $order->id : $order->get_id();
+		$old_wc   = version_compare( WC_VERSION, '3.0', '<' );
+		$order_id = $old_wc ? $order->id : $order->get_id();
 
-		if ( $order->has_status( array( 'processing', 'completed' ) ) ) {
+		if ( $order->has_status( array( 'completed' ) ) ) {
```
allendav: Why this change?

includes/class-wc-gateway-ppec-ipn-handler.php
```diff
@@ -183,7 +184,12 @@ protected function payment_status_completed( $order, $posted_data ) {
 			if ( ! empty( $posted_data['mc_fee'] ) ) {
 				// Log paypal transaction fee.
 				$transaction_fee = wc_clean( $posted_data['mc_fee'] );
-				update_post_meta( $order_id, 'PayPal Transaction Fee', $transaction_fee );
+				if ( $old_wc ) {
+					update_post_meta( $order_id, 'PayPal Transaction Fee', $transaction_fee );
```
allendav: Please DRY 'PayPal Transaction Fee'

includes/class-wc-gateway-ppec-admin-handler.php
marcinbot: This text is inspired by the Stripe plugin: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/b80e0075f550aebcce59b8753c0451f53a2d3dcb/includes/class-wc-gateway-stripe.php#L944

For example, I have a PayPal account set to use GBP, but my store is set up in USD. On PayPal, I can set up a holding in USD, but I can also opt in for the payments to be converted to GBP automatically. So I would be getting a payout in GBP, but wp-admin would show it in USD with a disclaimer that what I actually get may be in another currency.

includes/class-wc-gateway-ppec-ipn-handler.php
marcinbot: With my PayPal merchant settings, the payments would get automatically approved and the first response from PayPal would already be marked as completed: [class-wc-gateway-ppec-checkout-handler.php#L941](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/507c0ebc8a1b28d808c6518d739e8a06743a9cec/includes/class-wc-gateway-ppec-checkout-handler.php#L941). An IPN would then get sent containing the fee and more details and bounce back because of this. I think it's safe to remove here, because it only replicates the transaction ID being set on the order if it's already being processed.

Also, the initial PayPal response doesn't contain the customer data and is of different format than the IPN , but might contain a fee. I added processing of that fee in 507c0eb

includes/class-wc-gateway-ppec-ipn-handler.php
marcinbot: Done in 507c0eb

laurendavissmith:
@marcinbot i've updated the reviewer since @roykho is actually the reviewer right now, you can see who it is in our p2 sidebar

allendav:
@roykho  - are you sure you want to do that? From what I can tell, that meta key wasn't added by this PR. https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/510/files#diff-c37ad034021df5dedee3209418cc31c5L186

Wouldn't we have to migrate old transactions if we change it here?

roykho:
@allendav yes I am sure. However I thought we're just introducing this meta now hence didn't mention about migration. But yeah if we move forward with suggested changes, we will need to do a migration.

roykho:
So this is good, however I am not sure to what extent you want to migrate it. For example do you also need to migrate the DB for all the past orders? Or are we ok with just this. Otherwise this LGTM.

ChrisColotti:
I am on version 1.6.5 using PP Sandbox Testing and the fee / payout does not seem to be coming through.  Is this b/c I am using the PP sandbox?  I see this was added back in 1.4.3

ChrisColotti:
Thanks @marcinbot much appreciated!

Issue #517, opened 2019-01-15
10413/10426 Errors when Coupons are used


zwiesman31:
We're getting multiple reports on the forum about issues with version 1.6.6 when coupons of any kind are used: 

* https://wordpress.org/support/topic/paypal-error-10413-related-to-using-coupons/

`Error: PayPal error (10413): An error (10413) occurred while processing your PayPal payment. Please contact the store owner for assistance.`

From the logs: 

```
[L_ERRORCODE0] => 10413
[L_SHORTMESSAGE0] => Transaction refused because of an invalid argument. See additional error messages for details.
[L_LONGMESSAGE0] => The totals of the cart item amounts do not match order amounts.
[L_SEVERITYCODE0] => Error
```

Another user is reporting this issue when coupons are used: 

![http://cld.wthms.co/yfQbGh](http://cld.wthms.co/yfQbGh+)
Image link: http://cld.wthms.co/yfQbGh

The error comes up for customers after attempting to complete a payment (going through to PayPal and trying to complete the order/payment).

Reverting back to version `1.6.5` of [PayPal Checkout](https://woocommerce.com/products/woocommerce-gateway-paypal-checkout/) or switching to [PayPal Standard](https://docs.woothemes.com/document/paypal-standard/) helps resolve the issue.

zanza321:
I am currently having this error.   On my live site, I get the error when trying to use Paypal Express Checkout.  However, I setup a staging site (WP-Staging), and activated all plugins, and do not get the error.

So even with all plugins activated on staging site, there is no error.  However, one point, is on my staging site, I am not able to use USPS shipping from Woocommerce Services.  From what I read, new sites need to pay $79 to use the service, so I am not able to test if the bug is coming from this plugin.

keninbassart:
Reported in 1726324-Zen

PR #519, opened 2019-01-21
Fix Error 10413 when using coupons

DanReyLop:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/517
This bug was fixed in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/341, but it re-appeared in `1.6.6`. This PR fixes the `Error 10413` bug and also keeps the separate "Discount" lines in the PayPal flow.

I'm not sure how to test this since I haven't managed to reproduce it locally. It seems to be related with differences in precision when handling lots of numbers with decimals in the same order (shipping price, taxes, discount coupon, item price...). At the end, somehow the sum of all the parts is different than the total so PayPal fails.

@zanza321 @beehavendesignz I would prefer to be able to reproduce this error at least once before releasing the fix. I'm confident this won't break things more but we shouldn't just update the plugin without being sure that it works as intended. If you could tell us this information for the orders that failed, it would be extremely helpful:
- What option do you have in the `Subtotal Mismatch Behavior` setting of PayPal Checkout?
- Currency
- Product price
- Coupon price (or %). What kind of discount is it? Per-product, or per-cart?
- Shipping price for that order
- Tax amount for that order

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Changelog ***
 
+= 1.6.7 - 2019-XX-XX =
+* Fix - Error 10413 when using coupons
+
 = 1.6.6 - 2019-01-09 =
 * Fix - Discount items were not being included
 * Add - Filter for order details to accept decimal quantities of products
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -577,7 +577,7 @@ class WC_Gateway_PPEC_Client {
 		if ( $details['total_item_amount'] == $discounts ) {
 			// Omit line items altogether.
 			unset( $details['items'] );
-		} else if ( $discounts > 0 && $discounts < $details['total_item_amount'] ) {
+		} else if ( $discounts > 0 && $discounts < $details['total_item_amount'] && ! empty( $details['items'] ) ) {
 			// Else if there is discount, add them to the line-items
 			$details['items'][] = $this->_get_extra_discount_line_item($discounts);
 		}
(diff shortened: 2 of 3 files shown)


beehavendesignz:
Hello, sorry for me delay reply. I have downgraded back to version 1.6.5 and its working fine without this error. Here is the information I hope it is helpful.

What option do you have in the Subtotal Mismatch Behavior setting of PayPal Checkout? Do Not Send line items to PayPal
Currency is AUD
Product price 22.00
Coupon price (or %). What kind of discount is it? Per-product, or per-cart? Coupon fix price value $25.00
Shipping price for that order $2.50
Tax amount for that order $1.98

relwhild:
I am also in Australia and have similar problem - coupons will not work with paypal express checkout have had to disable:
What option do you have in the Subtotal Mismatch Behavior setting of PayPal Checkout? Do Not Send line items to PayPal
Currency is AUD
Product price 8.00
Coupon price (or %). What kind of discount is it? Per-product, or per-cart? Coupon percentage 20%
Shipping price for that order $4.95
Tax amount for that order $.58

zanza321:
I tried manually copy pasting the changes of the pull into my current version, just to test, and it caused some problems in the WP Admin backend.  Although, I did it quickly and possibly I made an error, but it seems to somewhat break Admin Woocommerce panel.  So, maybe I screwed it up but I wasn't able to test the fix .

 		} else if ( $discounts > 0 && $discounts < $details['total_item_amount'] && ! empty( $details['items'] ) ) {

zanza321:
yeah, I updated to new version, and its working properly now.  Seems to be fixed !

Issue #493, opened 2018-11-01
Javascript no longer enqueued as part of wp_enqueue_scripts


jenkoian:
Due to #491 the `wc-gateway-ppec-smart-payment-buttons` script no longer gets enqueued as part of the `wp_enqueue_scripts` hook, which means if theme developers are enqueuing JS that specifically depends on being later that this, it will no longer work if they're using `wp_enqueue_scripts` and instead they'll need to use a hook later in the process. This could come as a particular surprise to those not familiar with the hook lifecycle and that scripts could be loaded outside of `wp_enqueue_scripts`

That's a long way of saying this is a BC break and if following [semver](https://semver.org/) this should have gone in as a v2.x thing. Ideally though, you'd find a way to solve it whilst still enqueuing the JS in `wp_enqueue_scripts`.

DanReyLop:
> That's a long way of saying this is a BC break and if following semver this should have gone in as a v2.x thing.

Sorry to hear that this change affected your theme. Even though we don't officially follow `semver`, we try to avoid breaking changes as much as possible. I've never even considered #491 to be a breaking change until now.

> which means if theme developers are enqueuing JS that specifically depends on being later that this, it will no longer work if they're using wp_enqueue_scripts and instead they'll need to use a hook later in the process.

Let's try to fix that! :)
Just to clarify. Is your problem that you rely on your theme's JS to run *after* the PayPal script? Or are you trying to enqueue some extra JS **only if** the PayPal JS was enqueued?
Maybe we can add some hooks or JS events depending on what your problem is. Relying on the order on which JS is loaded seems quite fragile.

> Ideally though, you'd find a way to solve it whilst still enqueuing the JS in `wp_enqueue_scripts`.

I tried but that's simply not possible. A theme's template can render a "mini cart" in any page, so the only point when we can know for sure that the JS is needed is when the HTML for the PayPal Buttons is actually rendered, and that happens much later than `wp_enqueue_scripts`.

> This could come as a particular surprise to those not familiar with the hook lifecycle and that scripts could be loaded outside of `wp_enqueue_scripts`

In this case we could implement it this way because the PayPal JS script is loaded in the footer, so it can still be enqueued while the page's body is rendering.

jenkoian:
@DanReyLop thanks for your considered response! 😃 

I'll describe the issue we were facing and how we made it work and it may be that you have a better workaround anyway. 

[...]

PR #520, opened 2019-01-22
Always load the spb JS file, but only load the huge checkout.js file when needed

DanReyLop:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/493#issuecomment-452268356

After landing https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/491, PayPal SPB buttons stopped rendering in the mini-cart.

The problem is that the mini-cart HTML is cached in localStorage in the client, so the code that renders it in the server doesn't always run, so we can't rely on it for enqueuing the PayPal JS files.

I've implemented 2 alternative solutions, this one and #521. Please chime in on which one you prefer.

In this PR, I reverted the changes in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/491 (so that means that the `wc-gateway-ppec-smart-payment-buttons.js` is loaded on every page, but it's small enough), and as a compromise I made the **huge** `checkout.js` file load only when needed.

To test:
* Make sure that the Simple Payment Buttons load on the checkout, cart, product pages, and any page with the mini-cart.
* Check that our code only loads the `checkout.js` file once per page load.

### DIFF
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -144,19 +144,16 @@
 		}, selector );
 	};
 
-	var already_loading_script = false;
-	var renderWithLazyLoad = function( isMiniCart ) {
-		var onSuccess = render.bind( null, isMiniCart );
-		if ( already_loading_script ) {
-			onSuccess();
+	var load_script_task = null;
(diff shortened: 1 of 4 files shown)


RadoslavGeorgiev:
I'm closing this PR in favor of #521, which seems to me like the better solution.

PR #521, opened 2019-01-22
Render the PayPal buttons in the mini cart

DanReyLop:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/493#issuecomment-452268356

After landing https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/491, PayPal SPB buttons stopped rendering in the mini-cart.

The problem is that the mini-cart HTML is cached in localStorage in the client, so the code that renders it in the server doesn't always run, so we can't rely on it for enqueuing the PayPal JS files.

I've implemented 2 alternative solutions, this one and #520. Please chime in on which one you prefer.

In this PR, I made sure that the `wc-gateway-ppec-smart-payment-buttons` JS script is loaded whenever the mini-cart widget is rendered, by hooking into the `widget_title` hook.

To test:
* Make sure that the Simple Payment Buttons load on the checkout, cart, product pages, and any page with the mini-cart.

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -28,6 +28,7 @@ class WC_Gateway_PPEC_Cart_Handler {
 		} else {
 			add_action( 'woocommerce_widget_shopping_cart_buttons', array( $this, 'display_mini_paypal_button' ), 20 );
 		}
+		add_action( 'widget_title', array( $this, 'maybe_enqueue_checkout_js' ), 10, 3 );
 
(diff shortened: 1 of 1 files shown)


RadoslavGeorgiev:
Looks good and seems to work quite well. I like this solution better, so it is the one I choose :) :shipit:

RadoslavGeorgiev:
@dechov, yes, it does make a difference. Unless you want me to dig deeper, I will provide a shallow explanation:

1. I activated PPEC with Storefront.
2. Added products to the cart.
3. I did not see buttons on pages like the homepage of the website until I switched to this branch.

Issue #527, opened 2019-02-02
Smart Payment Buttons disables checkout


rossviviano:
Also reported [here](https://wordpress.org/support/topic/paypal-buttons-just-dissapeared/), when enabling the Smart Payment Buttons, the "Checkout with Paypal" button disappears from checkout. It also throws this error: 

`component.jsx:474 Uncaught TypeError: Cannot read property 'concat' of null
    at ParentComponent.decorate (component.jsx:474)
    at normalizeProp (props.js:44)
    at props.js:140
    at ParentComponent.setProps (index.js:440)
    at new ParentComponent (index.js:93)
    at index.js:488
    at Function.ZalgoPromise.try (promise.js:400)
    at Component.render (index.js:487)
    at callOriginal (util.js:386)
    at hacks.js:170`

This was reproduced on a site using Storefront and only WooCommerce & PayPal Checkout enabled.

Disabling the Smart Payment Buttons will allow customers to check out.

politeshrimp:
Sam issue here, please fix this.

sallyinstc:
Setting to horizontal makes the PayPal and PayPal Credit buttons show, but not the other credit card options. Also, I unchecked the Enable PayPal Credit checkbox, so that should not be showing.

dechov:
Thank you all for reporting the issue and offering workarounds – version 1.6.9 has now been released with a fix, and I've opened https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/529 to track the question of preventing such unexpected breaking incompatibilities.

Areahints:
Hello, I'm having the same problem on the latest version of the plugin.
```
wc-gateway-ppec-smart-payment-buttons.js?ver=2.0.2:49 Uncaught TypeError: Cannot read property 'FUNDING' of undefined
    at getFundingMethods (wc-gateway-ppec-smart-payment-buttons.js?ver=2.0.2:49)
    at render (wc-gateway-ppec-smart-payment-buttons.js?ver=2.0.2:82)
    at HTMLBodyElement.dispatch (jquery.js?ver=1.12.4-wp:3)
    at HTMLBodyElement.r.handle (jquery.js?ver=1.12.4-wp:3)
    at Object.trigger (jquery.js?ver=1.12.4-wp:3)
    at Object.a.event.trigger (jquery-migrate.min.js?ver=1.4.1:2)
[...]

PR #528, opened 2019-02-03
Avoid SPB render error by tweaking 'allowed' funding methods' empty value

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/527

A change in a recent release of PayPal's checkout.js is incompatible with the `null` value we were passing to `paypal.Button.render` for the `allowed` funding methods, breaking the rendering of the buttons. Passing `undefined` as an alternative works fine.

Can test by loading the Cart page with Smart Payment Buttons and vertical layout enabled: in this branch the buttons should render normally, while in `master` there should be no buttons and a browser console error.

_Edit:_ behavior can also be confirmed on PayPal Checkout's [demo page](https://developer.paypal.com/demo/checkout/#/pattern/vertical): `allowed: null` doesn't work, but `allowed: undefined` does.

This change brings the render call closer in line with not including the unneeded property at all (i.e. only `disallowed` is supported for vertical layout) – it is a minimal fix to solve the urgent issue.

### DIFF
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -38,7 +38,7 @@
 	// Map funding method settings to enumerated options provided by PayPal.
 	var getFundingMethods = function( methods ) {
 		if ( ! methods ) {
-			return null;
+			return undefined;
 		}
 
 		var paypal_funding_methods = [];


thesacredpath:
@dechov, how soon should this show up in user's dashboards as an available update?

dechov:
@thesacredpath Should be up in the next 10-20 minutes.

thesacredpath:
Thanks a bunch. I'll let the user know.

thesacredpath:
For reference, this is the user's chat transcript: https://mc.a8c.com/support-stats/happychat/transcript.php?id=10096750

And here is the follow-up ticket: https://woothemes.zendesk.com/agent/tickets/1772955
username: salfordrlfcst
site: https://salfordrlfcst.com/

PR #531, opened 2019-02-07
Add CURRENCYCODE do capture_payment

zedomel:
The missing CURRENCYCODE in capture_payment cause a error after payment authorization has been accepted:
(
    [AUTHORIZATIONID] => 2YC66033TH9230743
    [TIMESTAMP] => 2019-02-07T11:41:21Z
    [CORRELATIONID] => 31a70f4754928
    [ACK] => Failure
    [VERSION] => 120.0
    [BUILD] => 000000
    [L_ERRORCODE0] => 10613
    [L_SHORTMESSAGE0] => Currency mismatch.
    [L_LONGMESSAGE0] => Currency of capture must be the same as currency of authorization.
    [L_SEVERITYCODE0] => Error
)

Adding the CURRENCYCODE to params fix the problem.

### DIFF
--- a/includes/class-wc-gateway-ppec-admin-handler.php
+++ b/includes/class-wc-gateway-ppec-admin-handler.php
@@ -162,6 +162,7 @@ class WC_Gateway_PPEC_Admin_Handler {
 
 				$params['AUTHORIZATIONID'] = $trans_id;
 				$params['AMT'] = floatval( $order_total );
+				$params['CURRENCYCODE'] = get_woocommerce_currency();
 				$params['COMPLETETYPE'] = 'Complete';
 
 				$result = wc_gateway_ppec()->client->do_express_checkout_capture( $params );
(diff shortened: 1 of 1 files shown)


includes/class-wc-gateway-ppec-admin-handler.php
```diff
@@ -162,6 +162,7 @@ public function capture_payment( $order_id ) {
 
 				$params['AUTHORIZATIONID'] = $trans_id;
 				$params['AMT'] = floatval( $order_total );
+				$params['CURRENCYCODE'] = get_woocommerce_currency();
```
allendav: I think this should be something like `$order->get_currency()`

zedomel: @allendav  it sounds much better, as it order currency in context. Thanks!

allendav: If you can push an update to this PR today with that change, that would be awesome. If not, I can make the change and commit it.

zedomel:
@allendav  thanks for suggestions! I think that now we can close the request.

thanks.

PR #539, opened 2019-03-13
Check for option before attempting to delete it when bootstrapping.

jenkoian:
Check the options at bootstrap time before deleting them. As
get_option() is cached it can provide a small performance boost, or at
least can protect against a query occurring per request un-necessarily.

### DIFF
--- a/includes/class-wc-gateway-ppec-plugin.php
+++ b/includes/class-wc-gateway-ppec-plugin.php
@@ -161,12 +161,15 @@ class WC_Gateway_PPEC_Plugin {
 			$this->_check_credentials();
 
 			$this->_bootstrapped = true;
-			if ( false !== get_option( 'wc_gateway_ppce_bootstrap_warning_message' ) ) {
-				delete_option( 'wc_gateway_ppce_bootstrap_warning_message' );
-			}
 
-			if ( false !== get_option( 'wc_gateway_ppce_prompt_to_connect' ) ) {
-				delete_option( 'wc_gateway_ppce_prompt_to_connect' );
+			if ( is_admin() ) {
+				if ( false !== get_option( 'wc_gateway_ppce_bootstrap_warning_message' ) ) {
+					delete_option( 'wc_gateway_ppce_bootstrap_warning_message');
+				}
+
+				if ( false !== get_option( 'wc_gateway_ppce_prompt_to_connect' ) ) {
+					delete_option( 'wc_gateway_ppce_prompt_to_connect' );
+				}
 			}
 		} catch ( Exception $e ) {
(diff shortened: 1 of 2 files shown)


WPprodigy:
From what I can tell, these options are only ever used in wp-admin to show notices. So along with this change, it could be beneficial to only check/update in the first place if `is_admin()`.

laurendavissmith:
We'll need to get a changelog entry made before merging this

includes/class-wc-gateway-ppec-plugin.php
```diff
@@ -161,8 +161,13 @@ public function bootstrap() {
 			$this->_check_credentials();
 
 			$this->_bootstrapped = true;
-			delete_option( 'wc_gateway_ppce_bootstrap_warning_message' );
-			delete_option( 'wc_gateway_ppce_prompt_to_connect' );
+            if ( false !== get_option( 'wc_gateway_ppce_bootstrap_warning_message' ) ) {
```
superdav42: Please use tabs instead of spaces and wrap these in a `is_admin()` check as @WPprodigy mentioned.

budzanowski:
I don't see how this improves performance. I think that it actually makes it worse. 
If the cache fails we still use a query: https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/option.php#L100 so that is not an argument.

@superdav42 what do you think?

WPprodigy:
> If the cache fails we still use a query

That's okay, since it will happen much less often. The cache will be hit 99% of the time, thus no query.

As it currently is, there is always a direct DB query on page load.

budzanowski:
If it is 99% it makes sense. 👍

superdav42:
We could probably rewrite this to not use options at all since the error states can be determined at any time and do not need to be save in the DB. That would be more involved

includes/class-wc-gateway-ppec-plugin.php
jenkoian: Oops, looks like my editor wasn't configured quite right. Have rectified and opened this PR to add an [editorconfig](https://editorconfig.org/) file https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/543

jenkoian:
Thanks everyone for the feedback, actions taken:

* Wrapped in `is_admin()` call
* Fixed formatting (opening subsequent PR to add editorconfig support [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/543))
* Added changelog entry (may need to change date of it depending on release)

Happy to look at moving away from options altogether in a future PR if you wish 👍

c-shultz:
Thanks for the PR! #547 builds on the idea here and totally removes the options, so I'll close this in favor of that.

PR #540, opened 2019-03-14
Fixed plugin bootstrap invalidates options cache on every page view.

sun:
Problem
* The plugin woocommerce-gateway-paypal-express-checkout unconditionally deletes options on every page view, which causes the whole options cache of all autoload options to be regenerated on every page view.

Proposed solution
1. Only delete the options if they exist.

### DIFF
--- a/includes/class-wc-gateway-ppec-plugin.php
+++ b/includes/class-wc-gateway-ppec-plugin.php
@@ -161,8 +161,12 @@ class WC_Gateway_PPEC_Plugin {
 			$this->_check_credentials();
 
 			$this->_bootstrapped = true;
-			delete_option( 'wc_gateway_ppce_bootstrap_warning_message' );
-			delete_option( 'wc_gateway_ppce_prompt_to_connect' );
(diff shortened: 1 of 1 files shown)


sun:
This is a sister PR to https://github.com/woocommerce/woocommerce/pull/23035 (and some others outside of the WooCommerce universe)

sun:
Is there anything I can do to get this bugfix merged?

I assume that WordPress developers are aware that WordPress Core generates a cache of all options marked with `autoload = yes` in the database, so as to load those options faster on regular page views.

This cache is getting regenerated on every regular page view, because the woocommerce-gateway-paypal-express-checkout plugin unconditionally performs a delete operation on some options.

A regular page view should not trigger any write operations in the backend.

jenkoian:
I think this is a duplicate of #539 and has been addressed in #547

sun:
Yes, that's true. https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/547 replaces the options entirely with class properties. (There is not yet a plugin update to delete the obsolete options from the database though.)

PR #547, opened 2019-03-22
Remove option writes during bootstrap

WPprodigy:
It was pretty straightforward to just switch out the option writes for saving to internal values, as mentioned in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/539#issuecomment-474100787. These options were only used in one location, hooked onto `admin_notices`. This PR simply replaces them with class properties.

Testing:

The easiest way to test is to make manual edits in `_check_dependencies()` and `_check_credentials()` to just force the exceptions to be thrown / not thrown. Then just reload the admin and ensure the correct notices show up.

CC @laurendavissmith  / @superdav42

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** Changelog ***
 
+= 1.6.11 - 2019-xx-xx =
+* Fix/Performance - Prevent db option updates during bootstrap on each page load
+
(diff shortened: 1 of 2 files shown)


laurendavissmith:
Adding @c-shultz since he is PR reviewer this week

PR #548, opened 2019-03-28
Add hook on payment error.

jenkoian:
Should payment fail for whatever reason as well as throw an exception a
hook is added so that third parties can run any routines they need to in
this scenario.

Use case for this change is we want to be able to log on any payment
error (to new relic).

Stripe has a similar hook here: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/a23615803579646973ced040cb1d114d9f5a4032/includes/class-wc-gateway-stripe.php#L851

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 1.6.15 - 2019-* =
 * Fix - Prevent PHP errors when no billing details are present in PP response
 * Fix - Require billing address for virtual products when enabled
+* Add - Hook when a payment error occurs
 
 = 1.6.14 - 2019-05-08 =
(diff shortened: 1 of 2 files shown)


jenkoian:
No interest in this whatsoever?

jenkoian:
Absolutely no interest in this? It's such a minor thing to add but would seemingly offer a lot?

laurendavissmith:
@superdav42 assigning you as PR reviewer

@jenkoian apologies for the delay, thanks for the PR, do you mind adding a changelog entry for this PR as well

jenkoian:
Changelog entry added

laurendavissmith:
Thanks again for the PR!

PR #550, opened 2019-03-29
Check if the shipping address returned from PayPal is not empty

### LINKED ISSUE #549: Breaks when attempting to pay for a virtual product with WP_DEBUG enabled
To reproduce:
* Make sure `define( 'WP_DEBUG', true );` is present in `wp-config.php`
* Attempt to use the checkout button when purchasing any virtual product
* PayPal should not ask for shipping or billing address during checkout
* When PayPal redirects back to the site, a bunch of PHP errors will appear and order completion will not be possible

marcinbot:
Fixes #549 

Checks if the shipping address returned from PayPal is not empty before trying to use it to update the customer information.

To test:
* Make sure define( 'WP_DEBUG', true ); is present in wp-config.php
* Attempt to use the checkout button when purchasing any virtual product
* PayPal should not ask for shipping or billing address during checkout
* After returning to the site, no PHP notices should be shown and you should be able to complete the order

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -513,12 +513,14 @@ class WC_Gateway_PPEC_Checkout_Handler {
 		$customer->set_billing_state( $billing_details['state'] );
 		$customer->set_billing_country( $billing_details['country'] );
 
-		$customer->set_shipping_address( $shipping_details['address_1'] );
-		$customer->set_shipping_address_2( $shipping_details['address_2'] );
-		$customer->set_shipping_city( $shipping_details['city'] );
-		$customer->set_shipping_postcode( $shipping_details['postcode'] );
-		$customer->set_shipping_state( $shipping_details['state'] );
-		$customer->set_shipping_country( $shipping_details['country'] );
+		if ( ! empty( $shipping_details ) ) {
+			$customer->set_shipping_address( $shipping_details['address_1'] );
+			$customer->set_shipping_address_2( $shipping_details['address_2'] );
+			$customer->set_shipping_city( $shipping_details['city'] );
+			$customer->set_shipping_postcode( $shipping_details['postcode'] );
+			$customer->set_shipping_state( $shipping_details['state'] );
+			$customer->set_shipping_country( $shipping_details['country'] );
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -513,12 +513,14 @@ private function update_customer_addresses_from_paypal( $token ) {
...
 		$customer->set_billing_country( $billing_details['country'] );
 
-		$customer->set_shipping_address( $shipping_details['address_1'] );
-		$customer->set_shipping_address_2( $shipping_details['address_2'] );
-		$customer->set_shipping_city( $shipping_details['city'] );
-		$customer->set_shipping_postcode( $shipping_details['postcode'] );
-		$customer->set_shipping_state( $shipping_details['state'] );
-		$customer->set_shipping_country( $shipping_details['country'] );
+		if ( ! empty( $shipping_details ) ) {
+			$customer->set_shipping_address( $shipping_details['address_1'] );
```
allendav: Are we certain all these keys will be present? Would this be a good place to use wp_parse_args to make sure each key has a default?

includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -981,12 +983,14 @@ public function maybe_add_shipping_information( $packages ) {
...
 		$destination = $this->get_mapped_shipping_address( $checkout_details );
 
-		$packages[0]['destination']['country']   = $destination['country'];
-		$packages[0]['destination']['state']     = $destination['state'];
-		$packages[0]['destination']['postcode']  = $destination['postcode'];
-		$packages[0]['destination']['city']      = $destination['city'];
-		$packages[0]['destination']['address']   = $destination['address_1'];
-		$packages[0]['destination']['address_2'] = $destination['address_2'];
+		if ( ! empty( $destination ) ) {
+			$packages[0]['destination']['country']   = $destination['country'];
```
allendav: Same concerns as above. Are we certain all these keys will be present? Would this be a good place to use wp_parse_args to make sure each key has a default?

includes/class-wc-gateway-ppec-checkout-handler.php
marcinbot: Yeah, all of the keys are added by `$this->get_mapped_shipping_address`, which in turn parses the PayPal API response. At this point here we can be certain that the keys will be present.

If the API ever changes or causes parse errors for any reason, we should handle it as a separate issue.

PR #552, opened 2019-04-02
Better handling of virtual subs when billing address is not required

### LINKED ISSUE #545: Claims support for automatic renewals, even if requirements aren't met
## Synopsis

In *WooCommerce > Settings*, PayPal Checkout claims to support automatic recurring payments, even if 'billing address required' isn't checked.

If Reference Transactions aren't enabled, then subscriptions will be set to Manual Renewal, not automatic renewals.

If 'billing address required' isn't checked, then it's a much bigger issue, because subscription renewals will fail. 

## Observed behavior

Merchants are being led to believe that when they activate PayPal Checkout, automatic renewals will work in their store. It doesn't, unless they jump through two hoops which require communication and approval by PayPal. This is resulting in merchants selling lots of subscriptions, and then only realizing a month later that all the renewals are failing.

To understand that those things are required, they need to navigate to the documentation and find [this FAQ](https://docs.woocommerce.com/document/paypal-express-checkout/#section-17). This is clearly not enough, as we are getting hit with daily tickets with confused customers.

## Desired behavior

We need to prevent leading them into this trap. If PayPal Checkout doesn't support automatic renewals, we shouldn't be letting them think that it does. 

That should not show that it supports automatic recurring payments if billing address isn't required, and possibly also if Reference Transactions is not enabled.

Another possibility would be to show some large, unavoidable warnings if they have Woo Subscriptions and PayPal Checkout installed simultaneously.

## Steps to reproduce

1. Set up a test site with PayPal Checkout enabled, in sandbox mode
2. Verify the 'require billing address' field is not checked in *WooCommerce > Settings > Payments > PayPal Checkout*
3. Purchase a subscription product, verify that none of the address fields are filled out in the subscription

4. Manually [process renewal](https://docs.woocommerce.com/document/subscriptions/add-or-modify-a-subscription/#section-15)
5. See that the renewal fails because no addresses are returned to Woo to fill out the subscription 
```
03-20-2019 @ 05:58:44 - WC_Gateway_PPEC_Client::_process_response: acknowleged response body: Array
(
    [TIMESTAMP] => 2019-03-19T19:28:44Z
    [CORRELATIONID] => ef0bd95035e75
    [ACK] => Failure
    [VERSION] => 120.0
    [BUILD] => 46457558
    [L_ERRORCODE0] => 10727
    [L_ERRORCODE1] => 10728
    [L_ERRORCODE2] => 10004
    [L_SHORTMESSAGE0] => Invalid Data
    [L_SHORTMESSAGE1] => Invalid Data
    [L_SHORTMESSAGE2] => Invalid Data
    [L_LONGMESSAGE0] => There's an error with this transaction. Please enter an address1 in the shipping address.
    [L_LONGMESSAGE1] => There's an error with this transaction. Please enter a city in the shipping address.
    [L_LONGMESSAGE2] => This transaction cannot be processed.
    [L_SEVERITYCODE0] => Error
    [L_SEVERITYCODE1] => Error
    [L_SEVERITYCODE2] => Error
)
```

(note that there is some terminology confusion with PayPal for billing and shipping addresses, and if an address is collected and returned, it will fill out the billing address fields, and allow successful renewals)

marcinbot:
Fixes #545 

Modifies the Subscriptions tooltip if "Require billing address" option is not set:
Screenshot 2019-04-02 at 12 15 01

Removes PPEC from the available payment methods if the above option is not set and if the cart contains any virtual subscriptions.

To test:
* Verify that the tooltip appears on the settings page
* Create a virtual subscription and a shippable subscription products
* Attempt to checkout the virtual subscription - PPEC buttons should not appear at any point during the checkout
* Attempt to checkout the shippable subscription - PPEC buttons should appear
* Enable the "Require billing address" setting (for testing purposes I edited `woocommerce_ppec_paypal_settings` in `wp_options` in the database)
* The tooltip of the settings page should not show any changes
* Attempting to purchase a virtual subscription should show the PPEC buttons

### DIFF
--- a/includes/class-wc-gateway-ppec-with-paypal-addons.php
+++ b/includes/class-wc-gateway-ppec-with-paypal-addons.php
@@ -39,6 +39,73 @@ class WC_Gateway_PPEC_With_PayPal_Addons extends WC_Gateway_PPEC_With_PayPal {
 
 		add_action( 'woocommerce_scheduled_subscription_payment_' . $this->id, array( $this, 'scheduled_subscription_payment' ), 10, 2 );
 		add_action( 'woocommerce_subscription_failing_payment_method_' . $this->id, array( $this, 'update_failing_payment_method' ) );
+		add_filter( 'woocommerce_payment_gateways_renewal_support_status_html', array( $this, 'subscription_tooltip' ), 10, 2 );
+		add_filter( 'woocommerce_available_payment_gateways', array( $this, 'get_available_gataways_for_subscriptions' ), 20 );
+	}
+
+	/**
+	 * Filter for Subscriptions info tooltip html for this gateway
+	 *
+	 * @since 1.6.11
+	 *
+	 * @param string             $html HTML of the tooltip
+	 * @param WC_Payment_Gateway $gateway Payment gateway to filter for
+	 *
+	 * @return string Filtered HTML
+	 */
+	public function subscription_tooltip( $html, $gateway ) {
+		if ( $gateway->id !== $this->id ) {
+			return $html;
+		}
+		if ( 'no' === $gateway->get_option( 'require_billing', 'no' ) ) {
+			$tool_tip = esc_attr__( 'You must enable the "Require billing address" option to support this gateway\'s features for virtual subscriptions.', 'woocommerce-gateway-paypal-express-checkout' );
+			$status = esc_html__( 'Maybe', 'woocommerce-gateway-paypal-express-checkout' );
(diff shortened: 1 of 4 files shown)


phillipcornwell:
@allendav, @marcinbot: User from [1935264-zen](https://woothemes.zendesk.com/agent/tickets/1935264) says this is still removing the billing address with all cart items are virtual. Says that setting the checkbox doesn't fix this either.

PR #555, opened 2019-04-10
Filter button display on a per-product basis. Closes #554

### LINKED ISSUE #554: [Name Your Price compatibility] Filter to hide PayPal buttons for specific products
The quick pay buttons on a single product page don't work well with name your price products as they don't send the inputted price value.

In the short term, it's necessary to hide the buttons and force a user to use the cart/checkout in order to pay.

Could use a filter in the `WC_Gateway_PPEC_Cart_Handler::display_paypal_button_product()` method to disable the buttons on a per-product basis.

helgatheviking:
Closes #554 

Can be used like so:

```
/**
 * Disable PayPal buttons on NYP products.
 */
function kia_disable_paypal_buttons_for_nyp( $hide, $product ) {
	global $product;
	if( class_exists( 'WC_Name_Your_Price_Helpers' ) && ( WC_Name_Your_Price_Helpers::is_nyp( $product ) || WC_Name_Your_Price_Helpers::has_nyp( $product ) ) ) {
		$hide = true;
	}
	return $hide;
}
add_filter( 'woocommerce_paypal_express_checkout_hide_button_on_product_page', 'kia_disable_paypal_buttons_for_nyp', 10, 2 );
```

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -269,9 +269,10 @@ class WC_Gateway_PPEC_Cart_Handler {
 	 * @since 1.4.0
 	 */
 	public function display_paypal_button_product() {
+		global $product;
 		$gateways = WC()->payment_gateways->get_available_payment_gateways();
 
-		if ( ! is_product() || ! isset( $gateways['ppec_paypal'] ) ) {
+		if ( ! is_product() || ! isset( $gateways['ppec_paypal'] ) || apply_filters( 'woocommerce_paypal_express_checkout_hide_button_on_product_page', false, $product ) ) {
 			return;
 		}
 


allendav:
@helgatheviking - I'm curious - since `$product` is a global, why pass it in the filter?

includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -269,9 +269,10 @@ protected function set_customer_data( $data ) {
 	 * @since 1.4.0
 	 */
 	public function display_paypal_button_product() {
+		global $product;
 		$gateways = WC()->payment_gateways->get_available_payment_gateways();
 
-		if ( ! is_product() || ! isset( $gateways['ppec_paypal'] ) ) {
+		if ( ! is_product() || ! isset( $gateways['ppec_paypal'] ) || apply_filters( 'woocommerce_paypal_express_checkout_hide_button_on_product_page', false, $product ) ) {
```
allendav: Since $product is a global, why pass it in the filter?

allendav:
If we simplify the filter to this (I'd prefer we not make the conditional even longer - let's break it out on a separate line and NOT pass the product global:

```
		if ( ! is_product() || ! isset( $gateways['ppec_paypal'] ) ) {
			return;
		}
		
		if ( apply_filters( 'woocommerce_paypal_express_checkout_hide_button_on_product_page', false ) ) {
			return;
		}

		$settings = wc_gateway_ppec()->settings;
```

Then plugins can do things like this to hide products selectively:

```
function hello_hide( $hide ) {
	if ( function_exists( 'is_product' ) && is_product() ) {
		global $product;
		$product_name = $product->get_name( 'edit' );
		if ( FALSE === stristr( $product_name, 'HIDE' ) ) {
			return $hide;
		} else {
			return true;
		}
	}

	return $hide;	
}

add_filter( 'woocommerce_paypal_express_checkout_hide_button_on_product_page' , 'hello_hide' );
```

Will this work for you use case too, @helgatheviking  ?

allendav:
Opened an alternative PR in #604 - one that doesn't require passing the product global. Closing this one in favor of that.

PR #556, opened 2019-04-11
add filter woocommerce_paypal_express_checkout_hide_button_on_product_page

### LINKED ISSUE #554: [Name Your Price compatibility] Filter to hide PayPal buttons for specific products

helgatheviking:
Closes #554.

Some products (for example: Name Your Price products) cannot support payment buttons on the page and for now they need to be hidden.


dechov:
Closing as duplicate of your earlier PR https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/555

PR #567, opened 2019-05-06
Prevent PHP errors when no billing details are present in PP response

### LINKED ISSUE #561: Trying to get property of non-object
After upgrading to :
WooCommerce PayPal Checkout Gateway: 1.6.11
WooCommerce: 3.5.7
WordPress: 5.1.1

Our customer reported she couldn't pay, but this can also be an issue of not having funds.

Having the following notices in debug.log:

[23-Apr-2019 07:48:45 UTC] PHP Notice:  Trying to get property 'billing_address' of non-object in /var/www/prod/wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-checkout-handler.php on line 252
[23-Apr-2019 07:48:45 UTC] PHP Notice:  Trying to get property 'first_name' of non-object in /var/www/prod/wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-checkout-handler.php on line 255
[23-Apr-2019 07:48:45 UTC] PHP Notice:  Trying to get property 'last_name' of non-object in /var/www/prod/wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-checkout-handler.php on line 255
[23-Apr-2019 07:48:45 UTC] PHP Notice:  Trying to get property 'business_name' of non-object in /var/www/prod/wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-checkout-handler.php on line 408

When there are issues I would prefer having an application error message than a PHP notice.

marcinbot:
(Partially) fixes #561 

The error described in the issue above happens when PayPal does not return any billing details of the customer to the site. It might be a result of an issue in the API, since we weren't able to reproduce it. But we should still prevent the site from blowing up with PHP notices and this PR does that.

To test:
* enable debug logging
* add `return false;` at line 454 in `includes/class-wc-gateway-ppec-checkout-details.php`
* attempt to checkout
* no php errors should be visible when returning from PayPal
* the checkout should still not be possible because the billing fields are required
* there should be a message in the logs `PayPal response did not include the payer billing details` and the response

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -246,6 +246,7 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			wc_add_notice( $e->getMessage(), 'error' );
 			return;
 		}
+		// check if the payer details have been set on the PP response before rendering them to prevent PHP errors
 		if ( empty( $checkout_details->payer_details ) ) {
 			return;
 		}
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -246,6 +246,9 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			wc_add_notice( $e->getMessage(), 'error' );
 			return;
 		}
+		if ( empty( $checkout_details->payer_details ) ) {
+			return;
+		}
 		?>
 		<h3><?php _e( 'Billing details', 'woocommerce-gateway-paypal-express-checkout' ); ?></h3>
 		<ul>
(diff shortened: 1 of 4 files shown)


includes/class-wc-gateway-ppec-checkout-handler.php
```diff
@@ -246,6 +246,9 @@ public function paypal_billing_details() {
 			wc_add_notice( $e->getMessage(), 'error' );
 			return;
 		}
+		if ( empty( $checkout_details->payer_details ) ) {
```
jrodger: A comment might help keep this function readable, just to be explicit about why we're doing this "extra" check first?

jrodger:
Tested following your steps.

Suppressing the warnings is worthwhile and the changes stop the UI looking broken (even if you still can't proceed).

This should be ready to merge if you could look at that one comment (and rebase now or when merging).

jrodger:
Nice, thanks. Looks ready to go.

PR #568, opened 2019-05-08
Fix - Failing checkout when no addons are used

marcinbot:
Fixes failing checkout for sites with no addons installed


jrodger:
LGTM :shipit:

PR #570, opened 2019-05-09
Require billing address for virtual products when enabled

### LINKED ISSUE #569: Verify the "require billing address" checkbox behaviour
Originally reported [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/552#issuecomment-490615607).

When the cart only contains virtual products, the address fields are marked as not required. This happens also when the "Require billing address" setting is on.

Confirm that we still need to not require the address when the setting is on, or if it's possible to keep them as required.

marcinbot:
Fixes #569 

The billing address fields should remain required when the corresponding option is set in the admin panel.

To test:
* Use a regular merchant account with no address required option set
* Add only virtual products to the cart
* Open the checkout page
* The billing address fields (street, city, postcode) should not be required: Screenshot 2019-05-09 at 15 47 48
* Use a merchant with the billing address option enabled (reach out if you need sandbox credentials for such an account) or manipulate the database to force the "Require Billing Address" option to true
* Ensure the option is set
* With virtual products only in the cart, open the checkout page
* The billing address fields should be required: Screenshot 2019-05-09 at 15 51 41
* Disabling the option should make the fields not required again

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 1.6.15 - 2019-* =
 * Fix - Prevent PHP errors when no billing details are present in PP response
+* Fix - Require billing address for virtual products when enabled
 
 = 1.6.14 - 2019-05-08 =
 * Fix - Failing checkout when no addons are used
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -128,7 +128,7 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			return $fields;
 		}
 
-		if ( method_exists( WC()->cart, 'needs_shipping' ) && ! WC()->cart->needs_shipping() ) {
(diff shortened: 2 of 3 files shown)


jobthomas:
Any updates on this?

antonio86itna:
Hi, hope is fast...

Il Sab 18 Mag 2019, 16:58 Marcin Bot <notifications@github.com> ha scritto:

allendav:
@RadoslavGeorgiev - would you review this as janitor next week? Thx!

yukikatayama:
Curious if this would then also fix the [sub renewals issue](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/576)?

NickGreen:
@yukikatayama I'm not sure if this solves the issue - we'd have to test. The problem with the subs renewal issue is that people are using the PayPal Checkout shortcut buttons, which circumvents the normal WooCommerce Checkout page, and so I don't think simply requiring those fields on the Checkout page will work - since the customer never encounters those fields when using the shortcut buttons.

Issue #573, opened 2019-05-16
Custom order status not handled correctly when Authorizing payment


peterfabian:
I think the issue raised in https://github.com/woocommerce/woocommerce/issues/23626 applies here as well? 

Please see related PR for an inspiration for a fix: https://github.com/woocommerce/woocommerce/pull/23634

DanReyLop:
Thanks for the heads-up! I applied the same fix to the PayPal extension in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/574 and the Stripe extension in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/870

PR #574, opened 2019-05-18
Make the capture / cancel-auth processes work even if the order is in a custom status

DanReyLop:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/573

See https://github.com/woocommerce/woocommerce/issues/23626 and https://github.com/woocommerce/woocommerce/pull/23634 for context.


allendav:
@RadoslavGeorgiev - would you review this too as janitor next week? Thx!

Issue #516, opened 2019-01-03
"Phone: Optional/Required" setting in Customizer (from core WC) should indicate if it's being overridden by PPEC "Require Phone Number" Setting


dphyled:
"Require Phone Number" setting in PayPal Checkout overrides the WooCommerce core setting in Customizer under **WooCommerce > Checkout**, however there is no way to tell when you are in Customizer. This can lead to confusion.

For example in the following situation: 
1. You have PayPal Checkout active, and have not enabled the "Require Phone Number" setting (default):
![Screenshot](https://d.pr/YmevEd+)
Screenshot direct link: https://d.pr/YmevEd
2. You want phone to be required, so you go to Customizer > WooCommerce > Checkout, and set the phone to required:
![Screenshot](https://d.pr/NMhgpK+)
Screenshot direct link: https://d.pr/NMhgpK
3. You do some testing, and the phone is still not required, but you are unsure why.

Adding a little note under the customizer field when PPEC is active would avoid this confusion.

TimBHowe:
Ran into the same issue on a client site. We have created a simple pull request to disable and override the "Require Phone Number" setting in the plugin in favor of the Customizer settings if it is set to "Hidden" or "Required".

Hope this helps.

TimBHowe:
I have created a new PR.

The Customizer allow you to set the phone field as "Hidden", "Optional" or "Required".

The PPEC plugin only allows it to be "Optional" or "Required".

In the case where it is "Required" on the WC Customizer if we let the PPEC plugin setting override it the user would need to set the PPEC plugin as "Required" as well to actually require the phone number. This can lead to confusion and make the WC Customizer unnecessary or the PPEC plugin setting to be redundant.

In the case where it is "Hidden" on the WC Customizer and "Required" in the PPEC plugin the PPEC plugin tries to set the nonexistent billing_phone field as required in the includes/class-wc-gateway-ppec-checkout-handler.php#L171 and when it notices it is not there it skips it. In this case it could be very confusing to users to allow then to override the Customizer settings with the PPEC settings.

The PPEC plugin should probably just get rid of the require phone setting and use the Customizer setting in it's place. I have created an alternative pull request to show this solution and I think it is the better option.

Remove the "Require Phone Number" setting from the woocommerce-gateway-paypal-express-checkout plugin in favor of using the setting in the Cusomizer > WooCommerce > Checkout setting.

[...]

PR #577, opened 2019-06-05
Update the Required Phone Number Settings to Check Woo Customizer Settings

TimBHowe:
Added some checks to the plugin to disable and override the "Require Phone Number" setting in the WooCommerce > Payment > PayPal Checkout and use the WooCommerce Customizer setting if the phone number is set to "Hidden" or "Required".

Issue #516

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -165,6 +165,11 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			return $billing_fields;
 		}
 
+		// Check the WooCommerce customizer settings first.
+		if ( 'hidden' === get_option( 'woocommerce_checkout_phone_field', 'required' ) || 'required' === get_option( 'woocommerce_checkout_phone_field', 'required' ) ) {
+			return $billing_fields;
+		}
+
 		$require_phone_number = wc_gateway_ppec()->settings->require_phone_number;
 
 		if ( array_key_exists( 'billing_phone', $billing_fields ) ) {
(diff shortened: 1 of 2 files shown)


TimBHowe:
@menakas I have updated the pull request to reflect the change you have requested, but I have some concerns.

The Customizer allow you to set the phone field as "Hidden", "Optional" or "Required".

The PPEC plugin only allows it to be "Optional" or "Required".

In the case where it is "Required" on the WC Customizer if we let the PPEC plugin setting override it the user would need to set the PPEC plugin as "Required" as well to actually require the phone number. This can lead to confusion and make the WC Customizer unnecessary or the PPEC plugin setting to be redundant.

In the case where it is "Hidden" on the WC Customizer and "Required" in the PPEC plugin the PPEC plugin tries to set the nonexistent `billing_phone` field as required in the [`includes/class-wc-gateway-ppec-checkout-handler.php#L171`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-checkout-handler.php#L171) and when it notices it is not there it skips it. In this case it could be very confusing to users to allow then to override the Customizer settings with the PPEC settings.

The PPEC plugin should probably just get rid of the require phone setting and use the Customizer setting in it's place. I have created an alterative pull request to show this solution and I think it is the better option.

Please review and let me know if you need me to make any other changes.

https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/692

TimBHowe:
https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/516#issuecomment-603646478

Issue #414, opened 2018-05-16
Added fees sent to PayPal as tax, not as line items (as happens in PP Standard)


slash1andy:
When you add a fee to the cart, and then checkout using PPEC, the fee is sent with the tax to PayPal, and can mess with accounting, etc. 

This is sent as an item when using PP Standard, but not PPEC. 

To replicate: I used this plugin to add the fee: https://woocommerce.com/products/payment-gateway-based-fees/

Then go through the checkout with logging enabled, and look at the logs to see how the fee is sent. 
This happens (according to the customer) with 2 other fee plugins. 

1158988-zen 

@bor0

bor0:
I've added this to our triage list for discussion.

bor0:
This seems to be a compatibility issue. We can use PP Standard as a reference to find out what's missing (possibly some filter or action).

bor0:
@InoPlugs would you have an idea of what needs to be implemented for this compatibility ticket?

InoPlugs:
@bor0 

Our plugin uses 2 hooks on checkout page and modifies the cart content:

woocommerce_cart_calculate_fees for product specific fees
woocommerce_calculated_total for fees on total cart value

and it uses the following code to get the selected payment gateway:

	```
		$posted_payment_gateway = WC()->session->chosen_payment_method;
			
				//	possible fix, if calculate_... has been called before session data has been initialised
			if( function_exists( 'is_checkout' ) && is_checkout() && defined( 'WOOCOMMERCE_CHECKOUT' ) )
			{
				if( isset( $_REQUEST['payment_method'] ) )
				{
					$posted_payment_gateway = $_REQUEST['payment_method'];
				}
			}
```

Recalculation is triggered via ajax, when checkout page is loaded or user changes the payment gateway.

Currently our plugin does not support your checkout buttons on product page or cart page.

jbeales:
I ran into this today for a client site. Having the confirmation E-mail from PayPal and from WooCommerce say vastly different things about taxes charged is not a good look.

I'm motivated to get it fixed - I'll see if I can get a PR together.

marcinbot:
Closed in #583, will ship with v1.6.18

PR #583, opened 2019-06-19
Fees as line items

jbeales:
Fix for #414, based on the way PP Standard adds fees as line items.

Do we have any test suite or anything that this should pass?


allendav:
Thanks for the PR! We'll take a look at this for our upcoming release.

Issue #587, opened 2019-06-27
Missing email when checking out without PayPal Account


mikkamp:
When paying using the Express Checkout method, such as through the Smart Payment Buttons or from the Cart. We get an option to login to a PayPal account or pay directly filling in all the card/billing details.

When we choose the later options and fill in all the fields including "Email" and then select the option not to create a PayPal account. When we are redirected back to the WooCommerce site to complete the payment the Billing Email address is missing. When we try to complete the checkout we get the error:

```
Billing Email address is a required field.
```

Here is the test with a digital product, in this case we don't need a shipping address, so only the name is passed on from PayPal.

![image](https://user-images.githubusercontent.com/11388669/60275236-835d9780-98f1-11e9-8144-e15c0b1467db.png)

And here is the same with a physical product (here we selected the option in PayPal to ship to billing address):

![image](https://user-images.githubusercontent.com/11388669/60275992-a5a3e500-98f2-11e9-85c6-95a86b62263b.png)

If we pay with a PayPal account in either of these two cases then the Email address is passed.

In each of these two cases only the name is passed to WooCommerce, if we look at the log files we can confirm this and see the Email address is not passed in the response from PayPal:

```
06-27-2019 @ 14:21:09 - WC_Gateway_PPEC_Client::_process_response: acknowleged response body: Array
(
    [TOKEN] => EC-AAA620914D243364Y
    [BILLINGAGREEMENTACCEPTEDSTATUS] => 0
    [CHECKOUTSTATUS] => PaymentActionNotInitiated
    [TIMESTAMP] => 2019-06-27T14:21:09Z
    [CORRELATIONID] => 999b59aa66fc
    [ACK] => Success
    [VERSION] => 120.0
    [BUILD] => 53072421
    [PAYERID] => AAA5JZAN8GGH4
    [PAYERSTATUS] => unverified
    [FIRSTNAME] => Woo
    [LASTNAME] => Test
    [COUNTRYCODE] => US
    [ADDRESSSTATUS] => Confirmed
    [CURRENCYCODE] => USD
    [AMT] => 10.55
    [ITEMAMT] => 10.00
    [SHIPPINGAMT] => 0.00
    [HANDLINGAMT] => 0.00
    [TAXAMT] => 0.55
    [INSURANCEAMT] => 0.00
    [SHIPDISCAMT] => 0.00
    [INSURANCEOPTIONOFFERED] => false
    [L_NAME0] => Digital
[...]

PR #589, opened 2019-07-09
Avoid broken confirmation screen edge case after 10486 redirect

dechov:
Fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/561

As described in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/561#issuecomment-508274372, there are some broken edge cases around visiting the Checkout screen after a redirect to PayPal. I've reproduced them specifically for the redirect upon receiving a 10486 error, and I'm not sure whether there are other ways to reproduce such issues and/or if all have been addressed, but this is meant to fix the main issue as I've been able to reproduce it.

Screen Shot 2019-07-09 at 1 43 44 PM

- https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/7a9877caead9e956317384933c127dfc8852f6a3 adds first name and last name to the fields copied from the shipping address, so that they have values even if the billing address hasn't come back from PayPal. However, this **doesn't help for virtual products**.
- https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/976141b3d30e7b810f341315d2418d4afba6067c don't hide "Billing details" on checkout page, partially reverting https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/567 – instead, check for data before displaying
- https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/8c66656d347240dce1b26664d21fe684fd43e479 shows an email field if no email value is present in the session, <s>which also partially addresses https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/587 – a high priority issue with an unknown cause (seemingly on PayPal's side) – by still allowing the customer to re-enter the email address and proceed</s> (_edit_: this may be resolved now).
- https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/0a52e16275285425ca178237518dbfab7338fc07 addresses case of going Back via browser button after redirect on Checkout screen, which would result in a cached Checkout form but hidden submit button due to JS toggle intended for Smart Payment Buttons.

Spent some time trying to address remaining cases but kind of got lost in a rabbit hole around how required fields are manipulated in this extension (incl. https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/426). I also realized there can be validation errors without matching input fields in the normal flow [[screenshot](https://cloudup.com/ct0KAzUVVei)] if company or address_2 fields are required (in Customizer » WooCommerce » Checkout).

Is there reason not to add first_name and last_name (not to mention address_2 and probably company) to `not_required_fields`? Otherwise we might want to conditionally display those form fields on the confirmation screen.

To test / reproduce in `master`:
- Add `'AMT' => '104.86'` to `get_do_express_checkout_params` to trigger 10486 error after continuing from modal and confirming payment
- Check out, either:
  - with PayPal from Single Product or Cart, then after redirect navigate to /checkout
  - from Checkout screen, then after redirect go Back to Checkout with the browser (bypassing Cancel mechanism)
- In `master`, the confirmation screen's Continue button triggers validation errors for Billing First name, Last name, and Email fields, which aren't available for input
  - Note: the PHP notices were fixed in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/567

Testing with "Require Phone Number" setting enabled.

### DIFF
--- a/assets/js/wc-gateway-ppec-order-review.js
+++ b/assets/js/wc-gateway-ppec-order-review.js
@@ -2,6 +2,7 @@
 	'use strict';
 
 	$( 'form.checkout' ).on( 'click', 'input[name="payment_method"]', function() {
+		// Avoid toggling submit button if on confirmation screen
 		if ( $( '#payment' ).find( '.wc-gateway-ppec-cancel' ).length ) {
 			return;
 		}
--- a/assets/js/wc-gateway-ppec-order-review.js
+++ b/assets/js/wc-gateway-ppec-order-review.js
@@ -2,6 +2,10 @@
 	'use strict';
 
 	$( 'form.checkout' ).on( 'click', 'input[name="payment_method"]', function() {
+		if ( $( '#payment' ).find( '.wc-gateway-ppec-cancel' ).length ) {
+			return;
+		}
+
 		var isPPEC = $( this ).is( '#payment_method_ppec_paypal' );
 		$( '#place_order' ).toggle( ! isPPEC );
 		$( '#woo_pp_ec_button_checkout' ).toggle( isPPEC );
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -246,6 +246,12 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			wc_add_notice( $e->getMessage(), 'error' );
 			return;
 		}
+
+		if ( version_compare( WC_VERSION, '3.0', '<' ) ) {
+			$fields = WC()->checkout->checkout_fields['billing'];
+		} else {
+			$fields = WC()->checkout->get_checkout_fields( 'billing' );
+		}
 		?>
 		<h3><?php _e( 'Billing details', 'woocommerce-gateway-paypal-express-checkout' ); ?></h3>
 		<ul>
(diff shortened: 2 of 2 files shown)


allendav:
> adds first name and last name to the fields copied from the shipping address, so that they have values even if the billing address hasn't come back from PayPal. However, this doesn't help for virtual products.

Although not comprehensive (your point about virtual), this makes sense to at least grab what data we can when we have it.

> don't hide "Billing details" on checkout page, partially reverting #567 – instead, check for data before displaying

Makes sense. I'm curious why we only presenting billing address OR billing name but not both.

> shows an email field if no email value is present in the session

Makes sense.

> addresses case of going Back via browser button after redirect on Checkout screen

Makes sense.

allendav:
> Is there reason not to add first_name and last_name (not to mention address_2 and probably company) to not_required_fields? 

I agree. Looks like those fields could have / should have been added here: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blame/1a1f0544ea3b1e6bf01623157220e46f28e61b02/includes/class-wc-gateway-ppec-checkout-handler.php#L132

cc @bor0 - if you have any concerns about adding those fields to the "not required" array

assets/js/wc-gateway-ppec-order-review.js
```diff
@@ -2,6 +2,10 @@
 	'use strict';
 
 	$( 'form.checkout' ).on( 'click', 'input[name="payment_method"]', function() {
+		if ( $( '#payment' ).find( '.wc-gateway-ppec-cancel' ).length ) {
+			return;
```
allendav: Please add a comment explaining why we are doing this (i.e. https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/0a52e16275285425ca178237518dbfab7338fc07 ) so that maintainers don't have to run blame here

PR #599, opened 2019-08-06
Update for WooCommerce 3.7

allendav:
Test plan:

- [ ] Complete a test purchase using PayPal
- [ ] Complete a test purchase using a Credit Card
- [ ] As a US buyer, make sure Venmo is offered as a payment method to mobile buyers
- [ ] As a US buyer, make sure PayPal Credit is offered as a payment method
- [ ] Make sure you can authorize then capture
- [ ] Test with Smart Payment Buttons enabled (and not) (e.g. Venmo is one)
- [ ] Test with and without Checkout on Single Product
- [ ] Test with and without Enable PayPal Checkout on cart page
- [ ] Test with and without Enable PayPal Mark
- [ ] Test with and without an API Certificate
- [ ] Test with and without a PayPal Hosted Logo Image and Header Image
- [ ] Test with a tax rate that results in fractional cents and make sure purchases succeed
- [ ] Test with Subscriptions


allendav:
@marcin @dechov - just one review needed

allendav:
Thank you @dechov - I'm going to try to get #555 and #531 into the 1.6.17 release too, and if I do so, I'll update the changelog here before committing.

PR #613, opened 2019-08-22
Fixed cart value null issue.

yash-webkul:
Fixed cart value null issue.


mattallan:
hi @yash-webkul

Your changes introduce another issue because you have the `$needs_shipping` variable only being set inside the first `WC()->cart` condition and breaks the return statement

I have just pushed up #654 to address the issue and will close this PR. Thanks

PR #615, opened 2019-09-09
Add filter for PayPal redirect URL

dwardu:
Hi,

I have encountered a scenario where despite Guest Checkout being enabled, the customer was directed from the WooCommerce checkout to:

```
/checkoutnow?token=TOKEN&useraction=commit
```

which was then redirected to:

```
/checkoutnow?token=TOKEN&useraction=commit#/checkout/signup
```

instead of to `#/checkout/guest`, as was being expected since the PPEC plugin Landing Page had been configured to “Billing (Non-PayPal account)”.

On other occasions the customer was redirected correctly to `#/checkout/guest`, but an automatic client redirection to `#/checkout/signup` was triggered in the client as soon as the customer changed the credit-card country.

Through experimentation it was observed that setting the credit-card country to some countries, e.g. US/UK, always triggered a redirect to `#/checkout/guest`, whereas other specific countries, e.g. Malta, always triggered a redirect to `#/checkout/signup`.

In the end, the only reliable way I found to  work around this issue was to hardcode the `#/checkout/guest` fragment in the PPEC plugin.

By exposing the `woocommerce_paypal_express_checkout_paypal_redirect_url` filter I am introducing in this PR, it would be possible to implement such a customization from outside the PPEC plugin.

Thank you.

### DIFF
--- a/includes/class-wc-gateway-ppec-settings.php
+++ b/includes/class-wc-gateway-ppec-settings.php
@@ -177,7 +177,7 @@ class WC_Gateway_PPEC_Settings {
 			$url .= '#/checkout/chooseCreditOffer';
 		}
 
-		return $url;
+		return apply_filters( 'woocommerce_paypal_express_checkout_paypal_redirect_url', $url );
 	}
 
 	public function get_set_express_checkout_shortcut_params( $buckets = 1 ) {


mattallan:
Thanks for opening this PR.

I'm a little reluctant to merge this PR for a couple of reasons:
1. This URL is only used when smart payment buttons are disabled (this is being enabled by default in 1.7, so this functionality will be deprecated next version)

2. PayPal is automatically redirecting customers to these checkout forms based on their billing country (for example, Customers from France will go to `#/checkout/signup` and, customers from the US got to `#/checkout/guest`).<br>
If you filter our URL from this extension to force `#/checkout/guest`, PayPal will always load the Guest checkout payment form which is what you want.. however from my testing, if you change the country on the PayPal form after being on `/guest` you will get redirected back to `#/checkout/signup` if that country doesn't support guest checkout<br>
There's likely a reason why PayPal is asking customers to signup from certain countries, so I don't think filtering this URL is the best solution for this problem. My suggestion would be to contact PayPal to see if they can redirect all users to the guest checkout form and maybe try to get a reason for them doing this.

I'm going to close this for now. If there's no solution on PayPals side of things, we can reconsider this for as long as we support the non-smart payment buttons option

PR #620, opened 2019-09-25
Add get_paypal_locales() method

strarsis:
@gedex, @bor0, @dechov: 
This PR adds a `get_paypal_locales` method for getting an array of supported PayPal locales.
This can be helpful other plugins that want to use the `woocommerce_paypal_express_checkout_paypal_locale` filter.
It also applies a filter on the locales array so it is possible to quickly change the array of supported locales (as PayPal could add new supported locales).


menakas:
Thanks for the PR @strarsis 

Closing this as it is better not to allow filtering the supported locales. The PayPal supported locales are fixed. Mapping locales is possible with #624

PR #624, opened 2019-09-27
Add mapping for some locales

strarsis:
Maps some WordPress locales to PayPal locales.

The current WordPress locales are listed here: https://make.wordpress.org/polyglots/teams/
The current PayPal locales are listed here: https://developer.paypal.com/docs/api/reference/locale-codes/#supported-locale-codes

https://github.com/strarsis/woocommerce-gateway-paypal-express-checkout/blob/59f4db3ab60cab33752d6fad543b83f2c6c9e9d5/includes/class-wc-gateway-ppec-settings.php#L51-L59

### DIFF
--- a/includes/class-wc-gateway-ppec-settings.php
+++ b/includes/class-wc-gateway-ppec-settings.php
@@ -48,6 +48,16 @@ class WC_Gateway_PPEC_Settings {
 		'zh_TW',
 	);
 
+	protected $_locales_map = array(
+		'de_DE_formal' => 'de_DE',
+		'ja'           => 'ja_JP',
+		'nb_NO'        => 'no_NO',
+		'nn_NO'        => 'no_NO',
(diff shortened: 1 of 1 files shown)


NZI:
You probably want to fix the tab vs space indents on this

strarsis:
@NZI: I replaced the spaces with tabs.

strarsis:
@allendav, @dechov, @mide22: Could someone try this fix and merge it?
This will fix the locale issue on some sites that use some specific locales.

strarsis:
Bump!

menakas:
Thanks for the PR @strarsis 
I see two issues here: 
1. The set of locales mapped is not complete.
2. `in_array` checks for an element in the values of an array, but we want to check for its presence in the keys. 

Please track progress in PR #658.

PR #637, opened 2019-11-18
Add a notice about legacy payment buttons deprecation

### LINKED ISSUE #578: Warn users that legacy payment buttons will be phased out in lieu of Smart Payment buttons
If a user has Smart Payment Buttons not enabled, display a notice warning that support for legacy payment buttons will be removed in an upcoming release.

Screen Shot 2019-06-05 at 12 32 55 PM

marcinbot:
Fixes #578 

We actually already had a notice implemented [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/commit/a8a8b56fe739b26f01ddb73420ff200c04c7fcd2):
Screenshot 2019-11-18 at 13 33 20
It was shown if:
* PPEC is enabled
* Smart Payment Buttons are disabled
* `wc_gateway_ppec_spb_notice_dismissed` option is not in the db or is not set to `yes`

I changed the notice to the following to emphasize that the buttons are going away in the next release:
Screenshot 2019-11-18 at 13 56 49
It is still shown if PPEC is enabled but the buttons are disabled, but now it is also red and not dismissible. The only way to get rid of it is to enable Smart Payment Buttons.

### DIFF
--- a/includes/class-wc-gateway-ppec-plugin.php
+++ b/includes/class-wc-gateway-ppec-plugin.php
@@ -234,7 +234,7 @@ class WC_Gateway_PPEC_Plugin {
 		}
 
 		$setting_link = $this->get_admin_setting_link();
-		$message = sprintf( __( '<p>PayPal&nbsp;Checkout with new <strong>Smart&nbsp;Payment&nbsp;Buttons™</strong> gives your customers the power to pay the way they want without leaving your site.</p><p>The <strong>existing buttons will be deprecated and removed</strong> in the <strong>next</strong> release. Upgrade to Smart&nbsp;Payment&nbsp;Buttons in the <a href="%s">PayPal&nbsp;Checkout settings</a>.</p>', 'woocommerce-gateway-paypal-express-checkout' ), esc_url( $setting_link ) );
+		$message = sprintf( __( '<p>PayPal Checkout with new <strong>Smart Payment Buttons™</strong> gives your customers the power to pay the way they want without leaving your site.</p><p>The <strong>existing buttons will be removed</strong> in the <strong>next release</strong>. Please upgrade to Smart Payment Buttons on the <a href="%s">PayPal Checkout settings page</a>.</p>', 'woocommerce-gateway-paypal-express-checkout' ), esc_url( $setting_link ) );
 		?>
 		<div class="notice notice-error">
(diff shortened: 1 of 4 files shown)


includes/class-wc-gateway-ppec-plugin.php
```diff
@@ -233,26 +233,13 @@ public function show_spb_notice() {
...
-						dismiss_action: "ppec_dismiss_spb_notice",
-						nonce: "<?php echo esc_js( wp_create_nonce( 'ppec_dismiss_notice' ) ); ?>"
-					} );
-				} );
-			} )( jQuery );
-			</script>
-			<?php
-		}
+		$setting_link = $this->get_admin_setting_link();
+		$message = sprintf( __( '<p>PayPal&nbsp;Checkout with new <strong>Smart&nbsp;Payment&nbsp;Buttons™</strong> gives your customers the power to pay the way they want without leaving your site.</p><p>The <strong>existing buttons will be deprecated and removed</strong> in the <strong>next</strong> release. Upgrade to Smart&nbsp;Payment&nbsp;Buttons in the <a href="%s">PayPal&nbsp;Checkout settings</a>.</p>', 'woocommerce-gateway-paypal-express-checkout' ), esc_url( $setting_link ) );
```
v18: Is there a reason we're using `PayPal&nbsp;Checkout` instead of `PayPal Checkout` (And same for other instances of `&nbsp;`)?

I think we can remove the `deprecated` language here since we are effectively deprecating the legacy buttons in this release. I also updated the text a bit:

`<p>PayPal Checkout with new <strong>Smart Payment Buttons™</strong> gives your customers the power to pay the way they want without leaving your site.</p><p>The <strong>existing buttons will be removed</strong> in the <strong>next release</strong>. Please upgrade to Smart Payment Buttons on the <a href="%s">PayPal Checkout settings page</a>.</p>`

v18:
Left a small comment on the notice copy. Other than that, :shipit:

includes/class-wc-gateway-ppec-plugin.php
marcinbot: I updated the copy:
Screenshot 2019-11-20 at 17 04 13

Re: `&nbsp;` - I think we just wanted to keep the brand names together when wrapping lines. I don't think it affects this copy that much, so I'm ok with removing them.

PR #641, opened 2019-11-21
Fix call to a member function has_session() on null

### LINKED ISSUE #580: Error: Call to a member function has_session() on null / class-wc-gateway-ppec-cart-handler.php:514
This error has been showing up for the last 2 releases.

/wp-content/plugins/woocommerce-gateway-paypal-express-checkout/includes/class-wc-gateway-ppec-cart-handler.php:514

marcinbot:
Fixes #580 

The only way in which I could reproduce this was to add the following filter:
```
add_filter( 'woocommerce_is_rest_api_request', function ( $a ) {
	return true;
} );
```

Technically it's possible to add a plugin that modifies the REST paths and leads to this. 

The fix is both to use the same method as WC core is using to check for frontend requests (if REST is detected, then session is set to `null`). The `is_rest_api_request ` method was set to public only in the recent versions of core, so then we fall back to checking if session is not empty/falsey.

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -509,13 +509,15 @@ class WC_Gateway_PPEC_Cart_Handler {
 	 * Creates a customer session if one is not already active.
 	 */
 	public function ensure_session() {
+		// TODO: this tries to replicate Woo core functionality of checking for frontend requests.
+		// It can be removed once we drop support for pre-3.5 versions.
 		$frontend = ( ! is_admin() || defined( 'DOING_AJAX' ) ) && ! defined( 'DOING_CRON' ) && ! defined( 'REST_REQUEST' );
 
 		if ( ! $frontend ) {
 			return;
 		}
 
-		if ( ! WC()->session->has_session() ) {
+		if ( ! empty( WC()->session ) && ! WC()->session->has_session() ) {
 			WC()->session->set_customer_session_cookie( true );
 		}
 	}
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 1.6.18 - 2019-x-x =
 * Fix - Send fees to PayPal as line items
 * Fix - Fix error 10426 when coupons are used
+* Fix - Call to a member function has_session() on null
 * Add - Notice about legacy payment buttons deprecation
 
 = 1.6.17 - 2019-08-08 =
--- a/readme.txt
+++ b/readme.txt
@@ -104,6 +104,7 @@ Please use this to inform us about bugs, or make contributions via PRs.
 = 1.6.18 - 2019-x-x =
 * Fix - Send fees to PayPal as line items
 * Fix - Fix error 10426 when coupons are used
+* Fix - Call to a member function has_session() on null
 * Add - Notice about legacy payment buttons deprecation
 
 = 1.6.17 - 2019-08-08 =


includes/class-wc-gateway-ppec-cart-handler.php
```diff
@@ -509,13 +509,14 @@ public function enqueue_scripts() {
 	 * Creates a customer session if one is not already active.
 	 */
 	public function ensure_session() {
-		$frontend = ( ! is_admin() || defined( 'DOING_AJAX' ) ) && ! defined( 'DOING_CRON' ) && ! defined( 'REST_REQUEST' );
+		$is_rest_request = is_callable( WC(), 'is_rest_api_request' ) ? WC()->is_rest_api_request() : defined( 'REST_REQUEST' );
```
v18: Do we need to update how we check for a REST API request, or is this for testing purposes?

v18:
The fix looks good to me. Left a comment on how we check for a REST API request. I had a hard time reproducing it, though, so I asked for more details from the original issue author.

v18:
Noting here that we're removing the check for whether the request is a "frontend" one. I wanted to document why we're doing this:

We were checking these things to see if a request was a "frontend" one:

`$frontend = ( ! is_admin() || defined( 'DOING_AJAX' ) ) && ! defined( 'DOING_CRON' ) && ! defined( 'REST_REQUEST' );`

Since we will now check  if `empty( WC()-> session )` first, we will know that a request is "frontend" or not.

Case 1) If the session is not set, then we do nothing since we can't set the customer session cookie anyway, whichever kind of request this is.

If it's a frontend request, we used to cause an error, and now we do nothing 👌
If it's not a frontend request, we used to do nothing, and now we do nothing 👌

_(For completeness's sake, note that in the first case we used to `return;`, but it does not matter in this context since the hook is not acting on a return value.)_

Case 2) If the session is set, that means that the session was set by WC, and those same checks for frontend were already done here:

https://github.com/woocommerce/woocommerce/blob/135e1bebc9093ac23b9d8a77ec7c09083a92ad4c/includes/class-woocommerce.php#L554

Here's the relevant stack portion:

```
WooCommerce->initialize_session (woocommerce/includes/class-woocommerce.php:756)
wc_load_cart (woocommerce/includes/wc-core-functions.php:2277)
WooCommerce->init (woocommerce/includes/class-woocommerce.php:555)
```

Here's the frontend check:

https://github.com/woocommerce/woocommerce/blob/135e1bebc9093ac23b9d8a77ec7c09083a92ad4c/includes/class-woocommerce.php#L304

Note that we were checking whether the constant was `REST_REQUEST` was defined instead of the checks that are shown above. We are changing this portion of the check. Since we are "duck-type" checking for the same "duck" (is it a rest request?) I'm comfortable making the change.

If it's a frontend request, we used to set the cookie, and now we do set the cookie 👌
If it's not a frontend request, we used to do nothing, and now we do nothing 👌

There are no other ways to set the session.

Looks good, 🚢

v18:
I thought about this some more, and we should double-check that all of the different checks are done in all versions of WC that we support, which is [WC 2.6](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/368523c0eca09dfa5140cc7e2c60c8e55f01d75f/woocommerce-gateway-paypal-express-checkout.php#L15).

I did some digging:

I found a PR included in WC 3.3 that added a check for cron requests:

https://github.com/woocommerce/woocommerce/commit/b84f9fa112c3bbb4d93cef612c9abb9a3ca93df4

The check for frontend requests _does_ include a check for cron, so there's not an issue here. But, the parent commit shows no check for rest requests when checking for frontend requests:

https://github.com/woocommerce/woocommerce/blob/d90b817/includes/class-woocommerce.php#L246

🤔 

We first started checking for rest requests in WC 3.50 with PR https://github.com/woocommerce/woocommerce/pull/20000

And indeed in the previous version, WC 3.4.7, we set the session after checking for a frontend request:

```
if ( $this->is_request( 'frontend' ) ) {
	// Session class, handles session data for users - can be overwritten if custom handler is needed.
	$session_class = apply_filters( 'woocommerce_session_handler', 'WC_Session_Handler' );
	$this->session = new $session_class();
	$this->session->init();
```

_Source: https://github.com/woocommerce/woocommerce/blob/141e20a18365c1a896407db734ba958e1270288a/includes/class-woocommerce.php#L473-L477_

But we don't check for a rest request:

```
case 'frontend':
	return ( ! is_admin() || defined( 'DOING_AJAX' ) ) && ! defined( 'DOING_CRON' );
```

_Source: https://github.com/woocommerce/woocommerce/blob/141e20a18365c1a896407db734ba958e1270288a/includes/class-woocommerce.php#L243_

This implies to me that we can't rely on all of the checks that we had before being fulfilled because the session is set.

v18:
> Core will perform the frontend check and instantiate the object.

Yes, Core performs the frontend check. In WC 3.7 the frontend check includes a check for a rest request, but in WC 3.4.7 the frontend check does not check if it's a rest request.

Maybe I'm not thinking this through correctly? Here's how I'm looking at it:

Say it's version WC 3.4.7 and 1.6.18 of the plugin, and we are in `ensure_session`:

1. We don't check whether or not it's a rest request, only if `WC()->session` is not empty
2. If `WC()->session` is not empty, we proceed.
3. At this point, even though `WC()->session` is not empty WooCommerce Core has not checked if it's a rest request or not
4. But, in the current version (1.6.17) of the plugin, we do check if it's a rest request before proceeding

So, the case is that it is a rest request and we still set the session cookie.

We should not be trying to set the cookie on a rest request. So it seems like there would be an issue using version 1.6.18 of this plugin and 3.4.7 of WC.

What do you think?

v18:
I think maybe what you're saying that if `WC->session` is not set, then we don't set the cookie. And in all rest requests `WC->session` is not set, so we never set the cookie in rest requests. I don't think I'm sure of that since a rest request check was never performed during the request up until that point.

v18:
nice, 🚢

PR #676, opened 2020-02-06
Move inline settings javascript to `ppec-settings.js`

jorgeatorres:
### Description
When the [PPEC gateway is initialized](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3c145c84a446df86f217f829c9064998685e6161/includes/abstracts/abstract-wc-gateway-ppec.php#L104-L106) it loads file [`includes/settings/settings-ppec.php`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3c145c84a446df86f217f829c9064998685e6161/includes/settings/settings-ppec.php), which returns the set of configuration settings for the gateway.
This file also [enqueues some inline javascript](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3c145c84a446df86f217f829c9064998685e6161/includes/settings/settings-ppec.php#L53-L248) in order to handle the hiding and showing of certain settings on the settings screen (for example: to hide Live credential fields when the "Sandbox" environment is selected, and viceversa).

Because it is unconditionally enqueued inline, this JS gets added to WC pages on the frontend (and even to all pages if using things like the mini-cart widget). Because the CSS IDs targeted by the code are very specific, this is probably fine most of the time, but it's unnecessary nevertheless. We also have an admin-side settings JS file (`ppec-settings.js`) which is not only not loaded on the frontend but also seems like a more appropriate location for this code.

This PR moves the inline JS to `ppec-settings.js`.

**Note:** I couldn't find a reason for this JS to be enqueued on the frontend, but I might've missed something. If that's the case, feel free to comment and/or close the PR.

### Steps to test
1. Checkout `master` or `1.7_dev`.
1. Navigate to a product page, or to any page (if using the Storefront theme).
1. View the source code for the page and confirm that there's an inline script for the PPEC admin settings page, which looks something like this ([click for full code](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3c145c84a446df86f217f829c9064998685e6161/includes/settings/settings-ppec.php#L53-L248)):
   ```js
   jQuery( function( $ ) {
           var ppec_mark_fields      = '#woocommerce_ppec_paypal_title, #woocommerce_ppec_paypal_description';
           var ppec_live_fields      = '#woocommerce_ppec_paypal_api_username, #woocommerce_ppec_paypal_api_password, #woocommerce_ppec_paypal_api_signature, #woocommerce_ppec_paypal_api_certificate, #woocommerce_ppec_paypal_api_subject';
   [...]

   ```
1. Checkout `no_inline_settings_js`.
1. Confirm that there's no inline JS for the admin settings.
1. Go to the plugin settings on the backend (WC > Settings > Payments > PayPal Checkout > Manage) and confirm that the settings page works correctly. For example:
   - Verify that toggling the "Environment" shows/hides the Live/Sandbox fields accordingly. 
   - Confirm that checking/unchecking cart/mini-cart/single product/checkout specific settings shows and/or hides the dependent fields.


mattallan:
Thanks for this PR @jorgeatorres! I was also scratching my head when I saw this JS being enqueued like this.

>  I couldn't find a reason for this JS to be enqueued on the frontend, but I might've missed something. If that's the case, feel free to comment and/or close the PR.

I can't see any reason for this either. I also didn't realise this was being loaded on the frontend (product pages etc) and I'm going to assume that wasn't intended.

I've tested this PR by making sure the scripts are no longer loaded on the frontend pages and the settings page is working as intended. I also double checked all the JS being moved is only related to the settings page - LGTM

---

Note: I've rebased this PR to include some extra JS which was merged into `1.7_dev` but then went missing. These lines specifically: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/660/commits/f3fcfd60ed55657b2d361983a4f4eb45f7720b44

Merging :)

PR #663, opened 2020-01-29
Do not display smart buttons for external products and grouped products

menakas:
Fixes #460 and #478 

The smart buttons are displayed in the single product page of External/affiliate products and grouped product pages. Clicking on the smart button gives an error.

Along the lines of #662 I added a check for the product being of type external or grouped before the smart buttons are displayed.

**Steps to Test**
1. Create an external product EP
2. View the product page. 
3. Apart from the button that is linked to the external site, the PayPal smart buttons are also displayed on master. On this branch, the smart buttons are not displayed.
4. Repeat the above process for a grouped product.

Closes #460 and #478

Issue #562, opened 2019-04-25
PayPal API errors 10426 and 10413 with "Number of Decimals" set to 4 and discount coupon applied


Moomboh:
**Summary**
When WooCommerce is configured to calculate prices with 4 decimals (woocommerce_price_num_decimals), applying a percentage discount coupon leads to `PAYMENTREQUEST_0_ITEMAMT` being sent with 4 decimals. This causes PayPal API errors [10426 ](https://developer.paypal.com/docs/classic/api/errors/#10426) and [10413](https://developer.paypal.com/docs/classic/api/errors/#10413).
Without a coupon applied there are no issues with using 4 decimals.

**Steps to reproduce**
1. Install WooCommerce and WooCommerce PayPal Express Checkout
2. Set Woocommerce Options in "WooCommerce->Settings"

    - In the "General" tab set "Number of Decimals" to 4
    - In the "Tax" tab set "Prices Entered With Tax" to "Yes, I will enter prices inclusive of tax"

3. Add this code to functions.php or custom plugin:
```php
//Sets parameter for price display to only show 2 decimals regardless of "Number of Decimals" option
function myslug_decimals_price_args($args)
{
    $args['decimals'] = 2;
    return $args;
}
add_filter('wc_price_args', 'myslug_decimals_price_args', 10, 1);
```
4. Create a "Percentage discount" coupon
5. Apply the coupon
6. Try to checkout via the PayPal Smart Button or in the checkout via PayPal

**Expected behaviour:**
The payment should work when a coupon is applied regardless of the "Number of Decimals" option. All values should be rounded to 2 or 0 decimals (depending on currency) before being sent to the PayPal API. 

**Actual behaviour**
`[PAYMENTREQUEST_0_ITEMAMT]` is being sent with 4 decimals. This causes PayPal API errors [10426 ](https://developer.paypal.com/docs/classic/api/errors/#10426) and [10413](https://developer.paypal.com/docs/classic/api/errors/#10413).
The $discounts variable is not rounded before it gets added to `round( WC()->cart->cart_contents_total, $decimals )` or respectively `round( $order->get_subtotal(), $decimals )` in
[WC_Gateway_PPEC_Client::_get_details_from_cart()](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/054e0f3cd1e03e249da009d2f121c1b689f4f815/includes/class-wc-gateway-ppec-client.php#L454)
and
[...]

PR #677, opened 2020-02-07
Do not add discounts to total item amount and cause line item amount offset

### LINKED ISSUE #485: Confusing calculation of coupon code on PayPal order confirmation
The coupon discount is first added on PayPal once and than subtracted twice. This is confusing.

Order
woocommerce_20181004_order

PayPal confirmation
woocommerce_20181004_paypal

Woocommerce 3.4.5
WooCommerce PayPal Checkout Gateway 1.6.4

menakas:
Fixes #485 

Description
---

Whenever a coupon is added to provide a discount, the Purchase details in the Paypal buyer account shows an extra item added "Line Item Amount Offset", equal to the discount amount and then eventually deducted as a Shipping Discount Amount: https://d.pr/i/unojiU

This extra line item amount offset was introduced to account for rounding variations between WC and PayPal, but this has been applying to all discounts, even if there is no rounding issue.

This total_item_amount includes the total of items and discounts and the difference is added as an extra line item. So, the first commit 8417ab5 does not include discounts in it (in both the cases of getting the details from the order and the cart).

The next commit a9cf763 makes appropriate modifications to the checks and conditions with the above change in mind.

The last commit 4bc0ba9 deducts the discounts via 'coupon' line item to calculate the rounded total.

Steps to Test
---
**Test A**

1. Create a product with a decimal part, say $89.99 ( from #663)
2. Create a 99% discount coupon.
3. Purchase the product via PayPal Smart button after applying the coupon.
4. On master, the purchase details show the extra line item and an extra discount. On this branch, it looks clean.

**Test B**( from #562 )
1. Set Woocommerce Options in "WooCommerce->Settings"

    - In the "General" tab set "Number of Decimals" to 4
    - In the "Tax" tab set "Prices Entered With Tax" to "Yes, I will enter prices inclusive of tax"

1. Add this code to functions.php or custom plugin:

        //Sets parameter for price display to only show 2 decimals regardless of "Number of Decimals" option
        function myslug_decimals_price_args($args)
        {
        $args['decimals'] = 2;
        return $args;
        }
       add_filter('wc_price_args', 'myslug_decimals_price_args', 10, 1);

1. Create a "Percentage discount" coupon
1. Apply the coupon
1. Checkout via the PayPal Smart Button

To further test, add more calculations like tax, shipping etc.. In all cases, there is no line item amount offset added on this branch, even if there is a discrepancy in rounding causing a Shipping discount of 0.01 applied(Why shipping discount and not rounding adjustment will be addressed in another PR): https://d.pr/i/HkITTR

Closes #485

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -580,11 +580,11 @@ class WC_Gateway_PPEC_Client {
 		// the difference.
 		$diff = 0;
 
-		if ( $details['total_item_amount'] != $rounded_total ) {
+		if ( $details['total_item_amount'] + $discounts != $rounded_total ) {
 			if ( 'add' === $settings->get_subtotal_mismatch_behavior() ) {
 				// Add line item to make up different between WooCommerce
 				// calculations and PayPal calculations.
-				$diff = round( $details['total_item_amount'] - $rounded_total, $decimals );
+				$diff = round( $details['total_item_amount'] + $discounts - $rounded_total, $decimals );
 				if ( abs( $diff ) > 0.000001 && 0.0 !== (float) $diff ) {
 					$extra_line_item = $this->_get_extra_offset_line_item( $diff );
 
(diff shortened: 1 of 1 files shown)


mattallan:
Thanks for taking care of this one @menakas! I've tested the above cases including random tax, shipping amounts and couldn't see any issues with this.

This is a big improvement to the order/payment details listed on PayPal when the cart contains coupons so it will be nice to get this in 1.7.0 🎉 

I just pushed up af9fe68 to fix a change you made in 2f7089d (I assume you were supposed to remove the discount instead of subtracting it? Please confirm I'm not missing something)

LGTM :+1:

---

`master` :x:



`issue_485` ✅

PR #684, opened 2020-02-18
Version 1.6.20

mattallan:
* Fix - Upgrade the plugin on plugins loaded rather than on plugin init. PR#682


mattallan:
1.6.20 successfulyl deployed to WP.org: https://plugins.trac.wordpress.org/browser/woocommerce-gateway-paypal-express-checkout

PR #692, opened 2020-03-08
Fix/516 remove phone setting

TimBHowe:
Remove the "Require Phone Number" setting from the woocommerce-gateway-paypal-express-checkout plugin in favor of using the setting in the Cusomizer > WooCommerce > Checkout setting.

We have included instructions and a UI directing users to set it in the Customizer.

Issue #516


mattallan:
Closing this in favour of #697 

The Required Phone Number setting is currently not working at intended and it should not be affecting fields on the regular checkout.

The expected behaviour is that this setting should only make the Billing Address and Phone fields required or optional on the PayPal Checkout confirm order form. This "confirm order form" is only used when the customer has clicked on the PayPal Buy Now button on the cart, mini-cart or product pages and has returned to your store.

PR #696, opened 2020-03-16
Use base64_decode on API certificate while creating Credential object

### LINKED ISSUE #565: If certificate is entered, there's no way to remove it
If certificate is entered, there's no way to remove it without manually removing the entry for `woocommerce_ppec_paypal_settings` in the database. 

### Reported in: 
1988758-zen

### Replication steps

1. Upload an image as a cert in the settings and save
1. PPEC will no longer work and there's no way to remove the "cert"

### LINKED ISSUE #584: Unable to connect to PayPal using certificate file
When uploading a certificate file to connect to PayPal we get the following message:
`Error: The API certificate is not valid.`

The reason it's not valid is because the certificate file is base64 encoded before it's being saved in the database, see: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/1.6.15/includes/abstracts/abstract-wc-gateway-ppec.php#L231

And the same code for the sandbox certificate: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/1.6.15/includes/abstracts/abstract-wc-gateway-ppec.php#L243

However any action where the certificate is read / processed it's used directly without passing it through base64 decode first. For example here: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/1.6.15/includes/abstracts/abstract-wc-gateway-ppec.php#L289

Since that fails it returns the error that the certificate is invalid. Since the certificate is already base64 encoded it seems unnecessary to do this again. I tested by removing the `base64_encode` function when saving the certificate. 

After doing so I was able to connect to PayPal and successfully complete a payment using a certificate file.

The workaround to using a certificate is to switching to a signature instead. However each PayPal account can only use either a signature or a certificate (not both). So when the same PayPal account is also connected to a different platform (such as Shopify which requires a certificate), we are unable to connect.

Affected ticket: 2124119-zen

menakas:
**Issue**: #584 

---
### Description

While storing an API Certificate file, it is base64 encoded, but while retrieving it for use, it is not base64 decoded and so there are errors while using API certificate. 

### Steps to test:

1. Log into the Sandbox seller account at sandbox.paypal.com
1. Click on the gear icon on the top right and click on Account Settings.
1. Click on Update against API Access
1. Click on Manage API credentials under NVP/SOAP API integration (Classic)
1. If you already have API Signature here, remove it and request for an API Certificate. 
1. Download the API certificate, username and password. 
1. Go to the PPEC settings page from WooCommerce settings.
1. Use the API username and password just obtained. Choose the API Certificate file and then save.
1. On master, you should get "API Certificate not valid" error but on this branch it is saved with a success message.
1. Go a step further and try to purchase a product using this set of credentials. On this branch, an order is successfully placed.

### Documentation

I want note here that the documentation specifies [here](https://docs.woocommerce.com/document/paypal-express-checkout/#section-4) against point 5 that API Signature 'and' Certificate are required. Probably needs 'or' to be used. In addition, combining it with Live API Subject being optional makes the it sound as though the Signature and Certificate are also optional. Probably requires a rewrite.

Closes #584
Closes #565

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -190,32 +190,31 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 			return __( 'No API certificate on file.', 'woocommerce-gateway-paypal-express-checkout' );
 		}
 
-		$cert = @openssl_x509_read( $cert_string ); // @codingStandardsIgnoreLine
-		$out  = '';
+		$cert      = @openssl_x509_read( $cert_string ); // @codingStandardsIgnoreLine
+		$cert_info = $cert ? openssl_x509_parse( $cert ) : null;
+		$output    = '';
 
-		if ( false !== $cert ) {
-			$certinfo = openssl_x509_parse( $cert );
-			if ( false !== $certinfo ) {
-				$valid_until = $certinfo['validTo_time_t'];
-				if ( $valid_until < time() ) {
-					// Display in red if the cert is already expired
-					$expires = '<span style="color: red;">' . __( 'expired on %s', 'woocommerce-gateway-paypal-express-checkout' ) . '</span>';
-				} elseif ( $valid_until < ( time() - 2592000 ) ) {
-					// Also display in red if the cert is going to expire in the next 30 days
-					$expires = '<span style="color: red;">' . __( 'expires on %s', 'woocommerce-gateway-paypal-express-checkout' ) . '</span>';
-				} else {
-					// Otherwise just display a normal message
-					$expires = __( 'expires on %s', 'woocommerce-gateway-paypal-express-checkout' );
-				}
+		if ( $cert_info ) {
+			$valid_until = $cert_info['validTo_time_t'];
 
-				$expires = sprintf( $expires, date_i18n( get_option( 'date_format' ), $valid_until ) );
-				$out = sprintf( __( 'Certificate belongs to API username %1$s; %2$s.', 'woocommerce-gateway-paypal-express-checkout' ), $certinfo['subject']['CN'], $expires );
+			if ( $valid_until < time() ) {
+				// Display in red if the cert is already expired
+				$expires = '<span style="color: red;">' . __( 'expired on %s', 'woocommerce-gateway-paypal-express-checkout' ) . '</span>';
+			} elseif ( $valid_until < ( time() - 2592000 ) ) {
+				// Also display in red if the cert is going to expire in the next 30 days
+				$expires = '<span style="color: red;">' . __( 'expires on %s', 'woocommerce-gateway-paypal-express-checkout' ) . '</span>';
 			} else {
-				$out = __( 'The certificate on file is not valid.', 'woocommerce-gateway-paypal-express-checkout' );
+				// Otherwise just display a normal message
+				$expires = __( 'expires on %s', 'woocommerce-gateway-paypal-express-checkout' );
 			}
+
+			$expires = sprintf( $expires, date_i18n( get_option( 'date_format' ), $valid_until ) );
+			$output = sprintf( __( 'Certificate belongs to API username %1$s; %2$s.', 'woocommerce-gateway-paypal-express-checkout' ), $cert_info['subject']['CN'], $expires );
+		} else {
+			$output = __( 'The certificate on file is not valid.', 'woocommerce-gateway-paypal-express-checkout' );
 		}
 
-		return $out;
+		return $output;
 	}
 
 	/**
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -576,4 +576,32 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 
 		return ob_get_clean();
 	}
+
+	/**
+	 * Gets the description for an environment's certificate setting.
+	 *
+	 * Includes information about the certificate on file and remove link.
+	 *
+	 * @param string $environment The environment. Optional. Can be 'live' or 'sandbox'. Default is 'live'.
+	 * @return string The HTML string for an environment's certificate including a remove link if one is on file.
+	 */
+	private function get_certificate_setting_description( $environment = 'live' ) {
+		if ( 'live' === $environment ) {
+			$credentials = wc_gateway_ppec()->settings->get_live_api_credentials();
+		} else {
+			$credentials = wc_gateway_ppec()->settings->get_sandbox_api_credentials();
+		}
+
+		// If we don't have a certificate credential return the empty certificate info.
+		if ( ! is_callable( array( $credentials, 'get_certificate' ) ) ) {
+			return $this->get_certificate_info( '' );
+		}
+
+		return sprintf(
+			'%1$s <a href="#" class="wc_ppec_remove_certificate" data-environment="%2$s">%3$s</a>',
+			$this->get_certificate_info( $credentials->get_certificate() ),
+			esc_attr( $environment ),
+			__( 'Remove', 'woocommerce-gateway-paypal-express-checkout' )
+		);
+	}
 }
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -95,7 +95,12 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 		// Image upload.
 		wp_enqueue_media();
 
-		wp_enqueue_script( 'wc-gateway-ppec-settings', wc_gateway_ppec()->plugin_url . 'assets/js/wc-gateway-ppec-settings.js', array( 'jquery' ), wc_gateway_ppec()->version, true );
+		$screen = get_current_screen();
+
+		// Only enqueue the setting scripts on the PayPal Checkout settings screen.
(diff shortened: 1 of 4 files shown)


james-allan:
@menaka were there any further changes you planned to make? Also were you able to test purchasing products with a certificate? When I've tried I've received an error: 

```
An error occurred while trying to connect to PayPal: cURL error 58: unable to use client certificate (no key found or wrong pass phrase?)
```

menakas:
@james-allan 
I did not face any issues while purchasing a product. 🤔

menakas:
>were there any further changes you planned to make?
I was wondering if we should fix #565 as well. But I am not sure what to display as the certificate file name is not stored.

james-allan:
> I was thinking of fixing #565 as well. But I am not sure what to display as the certificate file name is not stored.

Oh yeah, that's a good one to fit in here.

At the moment with the current changes, the certificate information is shown. I think that's ok since we don't have the file name.

Screen Shot 2020-03-20 at 5 05 35 pm

I was thinking for #565 we could add a remove button or link to the end of that information.

james-allan:
RE https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/696#issuecomment-601562699

I've added a couple of commits so when there's a certificate saved, you can now remove it.

Screen Shot 2020-03-23 at 4 51 30 pm

james-allan:
Your changes look good to me @menakas :)

includes/abstracts/abstract-wc-gateway-ppec.php
```diff
@@ -209,7 +209,7 @@ private function get_certificate_info( $cert_string ) {
 				}
 
 				$expires = sprintf( $expires, date_i18n( get_option( 'date_format' ), $valid_until ) );
-				$out = sprintf( __( 'Certificate belongs to API username %1$s; %2$s', 'woocommerce-gateway-paypal-express-checkout' ), $certinfo['subject']['CN'], $expires );
+				$out = sprintf( __( 'Certificate belongs to API username %1$s; %2$s.', 'woocommerce-gateway-paypal-express-checkout' ), $certinfo['subject']['CN'], $expires );
 			} else {
 				$out = __( 'The certificate on file is not valid.', 'woocommerce-gateway-paypal-express-checkout' );
```
menakas: Would you know what settings would bring this text to be displayed? I tried a few different steps, but I could not get this message.
1. After a certificate is successfully set, I removed it from the PayPal Account settings.
2. I requested for a new certificate.
3. Now, I saved the WC PPEC settings page. But I cannot see this message.

james-allan: Yeah, that's a good question. I thought it would be shown if there was a file but it was invalid. When I uploaded an invalid file it didn't show anything though. 

eg: 
Screen Shot 2020-03-31 at 10 46 45 am

This is because when there's an invalid file, it doesn't pass the original `openssl_x509_read()` check. 

Looking at the code, this message would be shown if the cert can be read by `openssl_x509_read()` but cannot be parsed by `openssl_x509_parse()`, I don't know what circumstance would lead to that though.

For that reason, I've rejigged this part of the function in 42f6a28. 

----

Regarding your attempts to get this message. This notice is only displayed if the certificate is not a certificate file or it couldn't be parsed. It doesn't show if it's invalid as a credential on the account.

menakas: Thanks for the clarification James.

james-allan:
> The expiry date on the paypal account is in PDT.

ah, good eye. In my PayPal account, it shows in Australia Eastern Daylight time. I don't know why it's shown in PDT for you. From what I can understand it's using the timezone based on [your merchant's location](https://user-images.githubusercontent.com/8490476/77976251-dc0a0700-733f-11ea-94b7-7698b7db0f3b.png). 

Screen Shot 2020-03-31 at 11 03 01 am

In the code, when the cert file is read it pulls the timestamp. For the file in the screenshot above the timestamp is 1680146997 (30 March 2023 03:29:57) so the date displayed in the settings is UTC.

I've fixed that in 72a68b2. The expiry is now displayed in the site's time and the timezone abbreviation is also included to make that clear. I've tested this for both offset based (+10) and location-based (Brisbane) timezone settings.

Screen Shot 2020-03-31 at 11 43 56 am

Screen Shot 2020-03-31 at 11 43 27 am

------

> When there is a signature and then a certificate is added, the signature remains.

hmm yeah, that's an interesting one. From my understanding, when you add a cert the signature is no longer used. I've attempted to fix that in 28e56d5 and 36e82da. 

When you upload a cert file, the signature field is hidden. If you remove it both fields are displayed and if you input a signature, the cert field is hidden. 

Have a go @menakas and let me know whether that works for you. :)

menakas:
Thanks @james-allan 
I find this so much more clear. There is no room for most of the confusions that may have come up earlier.

I have two more smaller non-blocking concerns.

1. When there is a certificate file saved, reloading the page causes the 'No API Certificate on file' message to be displayed as a flash before replacing it with 'Certificate ....' message. If you think it is fine, I am okay with it.
2. Do you think it would be good to mention somewhere that the Certificate input will come up if Signature is empty and vice versa? Either in the tooltip or as a message above Signature. When someone has a Signature and want to move over to use Certificate, it might appear as though they don't have an interface to input the Certificate. That the Signature field should be emptied for the Certificate input to come up may need to be mentioned somewhere.

menakas:
@james-allan 
Another issue that I found while looking at some page:
When I had taxes enabled and went to the Standard tax rates, the "Loading.." message appears.
https://d.pr/i/JHiw6v

james-allan:
> When there is a certificate file saved, reloading the page causes the 'No API Certificate on file' message to be displayed as a flash before replacing it with 'Certificate ....' message. If you think it is fine, I am okay with it.

That flicker is caused by the lag between page load and the javascript running. When you first load the page, the live credentials are shown, as soon as the javascript loads, it hides the irrelevant set of credential fields which causes the flicker. It may be possible to fix that, however, I don't think it's worth the effort tbh. 

> Do you think it would be good to mention somewhere that the Certificate input will come up if Signature is empty and vice versa?

Hmm I think that would help prevent people from getting stuck if it's not clear. I don't know of a good way to do that though. Putting it in the tooltip may not be obvious and a note above the signature is rather difficult to add. Given this is a rather rare thing (to change your API credentials) I think it's fine the way it is. Perhaps we can make that clear in our documentation ([here](https://docs.woocommerce.com/document/paypal-express-checkout/#section-4)) and link to it from the settings.

> When I had taxes enabled and went to the Standard tax rates, the "Loading.." message appears.

oh good catch. That's caused by the files error. I didn't realise the settings.js file loads on all setting pages. I've fixed that in 79f55fa

james-allan:
> Perhaps we can make that clear in our documentation (here) and link to it from the settings.

I've added that in this commit 4ab115c. While working on that I also noticed that the placeholder text for the signature field wasn't being displayed for sandbox credentials. In that commit, I also fixed that.

includes/abstracts/abstract-wc-gateway-ppec.php
```diff
@@ -95,7 +95,13 @@ public function enqueue_scripts() {
 		// Image upload.
 		wp_enqueue_media();
 
-		wp_enqueue_script( 'wc-gateway-ppec-settings', wc_gateway_ppec()->plugin_url . 'assets/js/wc-gateway-ppec-settings.js', array( 'jquery' ), wc_gateway_ppec()->version, true );
+		$screen = get_current_screen();
+
+		// Only enqueue the setting scripts on the PayPal Checkout settings screen.
+		if ( true ) {
```
menakas, line 101: Sorry, if this is obvious. I do not understand the reason for this condition!

menakas:
Thanks James.

It is a great idea to link to the documentation here. 

As there appears to be some minor changes required for clarity in the documentation, I'm adding the docs label here. I edited the PR description to reflect the issue so that it is easy to figure what change is required.

includes/abstracts/abstract-wc-gateway-ppec.php
james-allan: haha oopps. Yeah that was a mistake 🤦‍♂ I've rebased that.

menakas:
I have been thinking all along that all the three are required for the credentials to be valid.

I tried saving the certificate without username and password. I got the "Your settings have been saved" message. The same is true of Signature. One can save the Signature without the username or password. But when the username only is given, an error is thrown asking for password. This seems to be inconsistent behaviour. Your thoughts?

james-allan:
> This seems to be inconsistent behaviour. Your thoughts?

Yeah, it's inconsistent. I think there are a few problems here:

1. All the fields that are required aren't validated. You need to have a username, password and signature or certificate. At the moment, only the password is validated and it's only validated if a username is provided. This will need to be fixed in `validate_active_credentials()`. 
2. They aren't all validated together. If you just add a signature it saves just fine. Essentially if you add any field but leave the username empty. It just saves. 

These changes are fairly large and I'd recommend refactoring `validate_active_credentials()` in large parts. I'm going to open a new issue for that. I think the changes required would blow out this PR which is already getting pretty large.

menakas:
Thanks @james-allan for the explanation.
Another thanks for your patience in getting this to this state.
Merging...

james-allan:
> only the password is validated and it's only validated if a username is provided. 

Just making a note about this. This isn't quite correct. The signature and certificate are validated too but you need to have added a username and password. If you have added a username and password but don't add a certificate or signature, you get the notice: `Error: You must provide API signature or certificate.`

PR #704, opened 2020-03-27
Exit early from js scripts before targeting nonexistent elements

### LINKED ISSUE #558: In out-of-stock single product pages: Uncaught TypeError
This error does not affect the operation but it should be fixed. 

### Steps to replicate the issue 

1. Create an out-of-stock product.
2. Visit its single product page.
3. Check the browser console, this JS error displays: 

```
Uncaught TypeError: Cannot read property 'checkValidity' of undefined
    at validate_form (wc-gateway-ppec-generate-cart.js?ver=1.6.11:46)
    at wc-gateway-ppec-generate-cart.js?ver=1.6.11:70
    at wc-gateway-ppec-generate-cart.js?ver=1.6.11:126
```

Screen Shot 2019-04-18 at 13 47 29

### What expected

No error in the browser console. 

### Environment 

* Default WP theme.
* WooCommerce PayPal Checkout Gateway 1.6.11 and WooCommerce 3.6.1

### Addtional context 

https://wordpress.org/support/topic/uncaught-typeerror-checkvalidity-when-product-is-out-of-stock/

chickenn00dle:
Per issue #558, when on a single product page containing an item that is out of
stock, both `wc-gateway-ppec-smart-payment-buttons.js` and
`wc-gateway-ppec-generate-cart.js` scripts throw exceptions as both
scripts are trying to target elements which don't exist when the product is out of stock:

From `wc-gateway-ppec-generate-cart.js`:
```
TypeError: Cannot read property 'checkValidity' of undefined
```

From `wc-gateway-ppec-smart-payment-buttons.js`:
```
Error: Document is ready and element #woo_pp_ec_button_product does not exist
```

### Steps to reproduce

From `master`:

1. Create an out-of-stock product.
2. Visit its single product page.
3. Check the browser console/source tabs

### Description

This PR adds a check for the relevant form element in each script and returns early if it is not present in the single product view. Doing so stops the scripts before they target invalid elements and resolves above exceptions.

Fixes #558

### DIFF
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -2,14 +2,6 @@
 ;(function( $, window, document ) {
 	'use strict';
 
-	var form = $( 'form.cart' );
-
-	// Return early in cases where no form is present
-	// i.e. out of stock product
-	if ( ! form.length ) {
-		return;
-	}
-
 	// This button state is only applicable to non-SPB click handler below.
 	var button_enabled = true;
 	$( '#woo_pp_ec_button_product' )
@@ -44,6 +36,7 @@
 	// True if all the fields of the product form are valid (such as required fields configured by Product Add-Ons). False otherwise
 	var fields_valid = true;
 
+	var form = $( 'form.cart' );
 
 	var update_button = function() {
 		$( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -2,6 +2,14 @@
 ;(function( $, window, document ) {
 	'use strict';
 
+	var form = $( 'form.cart' );
+
+	// Return early in cases where no form is present
+	// i.e. out of stock product
+	if ( ! form.length ) {
+		return;
+	}
+
 	// This button state is only applicable to non-SPB click handler below.
 	var button_enabled = true;
 	$( '#woo_pp_ec_button_product' )
@@ -36,7 +44,6 @@
 	// True if all the fields of the product form are valid (such as required fields configured by Product Add-Ons). False otherwise
 	var fields_valid = true;
 
-	var form = $( 'form.cart' );
 
 	var update_button = function() {
 		$( '#woo_pp_ec_button_product' ).trigger( ( variation_valid && fields_valid ) ? 'enable' : 'disable' );
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -152,6 +152,11 @@
 
 	// Render cart, single product, or checkout buttons.
 	if ( wc_ppec_context.page ) {
+		// Return early in cases where no form is present in product page
+		// i.e. out of stock product
+		if ( 'product' === wc_ppec_context.page && ! $( 'form.cart' ).length ) {
+			return;
+		}
 		if ( 'checkout' !== wc_ppec_context.page ) {
 			render();
 		}
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -150,6 +150,16 @@
 		}, selector );
 	};
 
+	// Render buttons in mini-cart if present.
+	$( document.body ).on( 'wc_fragments_loaded wc_fragments_refreshed', function() {
+		var $button = $( '.widget_shopping_cart #woo_pp_ec_button_mini_cart' );
+		if ( $button.length ) {
+			// Clear any existing button in container, and render.
+			$button.empty();
+			render( true );
+		}
+	} );
+
 	// Render cart, single product, or checkout buttons.
 	if ( wc_ppec_context.page ) {
 		// Return early in cases where no form is present in product page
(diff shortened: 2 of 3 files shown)


mattallan:
Woot! Nice first PR @chickenn00dle ! 💃 

Looking at these changes and calling `return;` at the top of the page, I wonder if we should be looking more for a change in PHP to not enqueue `wc-gateway-ppec-generate-cart.js` to begin with.

For instance, inside [`display_paypal_button_product()`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-cart-handler.php#L289) we are only loading the PayPal Buttons JS file if the:
 - product is in stock
 - product is not of type external or grouped
 - filter `woocommerce_paypal_express_checkout_hide_button_on_product_page` hasn't been set to `true`

Given that, what if we move the [`enqueue('wc-gateway-ppec-generate-cart')`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3f0c5f5176bdd202b3220563ff9e1a1837b2a6c1/includes/class-wc-gateway-ppec-cart-handler.php#L518-L524) to [`display_paypal_button_product()`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/master/includes/class-wc-gateway-ppec-cart-handler.php#L289)? This might be considered a breaking change so alternatively, we could add some additional logic to `$is_product` to somehow check is in stock etc.

The other consideration I have is if we return/exit early or don't load the file, are there unrelated functions/event listeners that 3rd party devs might be using but won't exist now we have `returned;`?

If so, we could look at fixing these console errors by adding some safeguards to the JS to make sure we're checking for variables or elements before assuming they exist.

What are your thoughts given these two cases?

chickenn00dle:
Thanks for the detailed review @mattallan !

> What are your thoughts given these two cases?

Personally, I think the cleanest option is to move `enqueue('wc-gateway-ppec-generate-cart')` to `display_paypal_button_product()`. But I definitely wouldn't want to make a breaking change, so would go with adding safeguards if that is the case. 

Looking through the script for potential conflicts, I am seeing we are providing a custom event with which 3rd parties can add validation checks of their own:
https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/c47ab440b1eab5b8c1e2ab545476dc7ed15e6549/assets/js/wc-gateway-ppec-generate-cart.js#L54

I would imagine this becomes redundant if a store owner indicates a product is out of stock. 

It looks like we are also adding `generate_cart` to the window object:
https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/c47ab440b1eab5b8c1e2ab545476dc7ed15e6549/assets/js/wc-gateway-ppec-generate-cart.js#L119

But again, I think this becomes redundant if a product is out of stock.

If these are the only potential issues here, I think going with option 1 would be best. 

Would you mind having a look through and confirming whether I've missed something?

mattallan:
I don't think you've missed anything there 👍 

> I definitely wouldn't want to make a breaking change, so would go with adding safeguards if that is the case.

Yeah I agree. This would be my preference as well.

I would say it makes the most sense for this to be enqueued within `display_paypal_button_product()` along with the PayPal Smart Buttons JS and I want to say this has very little chance of breaking third party code. 

So I say let's go with this approach and I'll test it against other 3rd party plugins before merging 😄

chickenn00dle:
Awesome! 

I've pushed two commits for this. The first moves `enqueue('wc-gateway-ppec-generate-cart')` to `display_paypal_button_product()` and the second resolves a bug introduced by exiting early from `wc-gateway-ppec-smart-payment-buttons.js` in which the minicart buttons weren't being rendered prior to the script exiting.

mattallan:
I believe we're getting close! 🥳

I was looking into why we needed to add [these lines](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/704/files#diff-98c725c071bfe8a62a19e2e29ece8be5R165-R169) and why it broke the mini-cart and I'm not sure about checking for the`form.cart` element here and returning.

The real issue is our render function assumes the [ppec product page `selector`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/af5e87f7c214786279a7bbc6361829d18871d585/assets/js/wc-gateway-ppec-smart-payment-buttons.js#L62) exists on the page.

Should we look at removing the [`if ( 'product' === wc_ppec_context.page && ! $( 'form.cart' ).length )`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/704/files#diff-98c725c071bfe8a62a19e2e29ece8be5R165-R169) and instead updating the render function specifically [this condition](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/af5e87f7c214786279a7bbc6361829d18871d585/assets/js/wc-gateway-ppec-smart-payment-buttons.js#L65-L68) to return the selector doesn't exist.

i.e.
```
// Don't try render if selector doesn't exist or if already rendered in DOM
if ( ! $( selector ).length || $( selector ).children().length ) {
	return;
}
```

Thoughts?

chickenn00dle:
I like this approach much better. Again, this is much cleaner and makes perfect sense. I removed the `if ( 'product' === wc_ppec_context.page && ! $( 'form.cart' ).length )` check and went with your idea. 

Thanks for pointing this out!

mattallan:
This LGTM 😎 Awesome first PR @chickenn00dle ! 

---

> I'll test it against other 3rd party plugins before merging

Just making a note I've tested this PR with [Mix and Match products](https://d.pr/v/lwUWXL) as well as: 
 - [x] simple and variable products not in stock
 - [x] simple and variable products in stock
 - [x] standard checkout flow
 - [x] mini cart

No issues there :100: Merging!

chickenn00dle:
Thanks Matt! 🎉

PR #705, opened 2020-04-01
Check for error notices added via wc_add_notice during checkout validation

### LINKED ISSUE #686: Non owned coupon error is not shown when choosing PayPal
**Describe the bug**
When adding a non owned coupon with PayPal, there is no error shown before payment.

**To Reproduce**
Add a non owned coupon, which is restricted by mail address, it will be applied to the cart. When using another email address and choosing PayPal, it opens the PayPal window.
In PayPal the full price is shown. Only when the customer logs in to PayPal and cancels, he will see that the coupon could not be used.

**Expected behavior**
Before opening the PayPal window, the coupon should be validated (maybe this is done), and the error shown to the customer.

 - WordPress Version: 5.3.2
 - WooCommerce Version: 3.9.2
 - PayPal Express Checkout Plugin Version: 1.6.20

chickenn00dle:
Per #686, when validating during checkout, any errors added via
`wc_add_notice` are missed when checking for errors in [maybe_start_checkout()](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3f0c5f5176bdd202b3220563ff9e1a1837b2a6c1/includes/class-wc-gateway-ppec-cart-handler.php#L173). As a
result, the PayPal flow is not interrupted in cases where an invalid
(unowned) coupon is entered and the relevant error notice appears only AFTER a user completes the PP process.

### Steps to reproduce:

1. Create a coupon restricted by email (**WooCommerce > Coupons > Add Coupon**)
2. From the stores front end, add an item to cart, and go to checkout
3. Be sure to enter an email that is not valid for the coupon in step 1
4. Select PP button
5. After being directed through the PP checkout flow, you will then see an error indicating the coupon is not yours.

### Description

This adds a check for these error notices, then merges them with
WP_Error object errors if they exist.

This also forces the checkout form to submit on error in order to
enforce any error behavior, i.e. remove invalid coupons

Fixes #686


omar09091993 (approved):

menakas:
@chickenn00dle 
I did some basic testing. Nice work.

While the error message is shown in the Checkout page, it is possible to go to the paypal modal by clicking on the smart payment button from the cart page. I think this should also give a notice.
Your thoughts?

chickenn00dle:
Thanks for looking over this @menakas! 

I'm not sure if this is possible from the cart page. At this point, the user would not have entered any billing details from which to compare the coupon email validation.

And forcing the user to provide an email address when entering an email restricted coupon is outside the scope of PPEC I think.

I may be missing something though, so if you have any suggestions on finding a way to validate the coupon from the cart page, please share!

menakas:
@chickenn00dle 
>I'm not sure if this is possible from the cart page. At this point, the user would not have entered any billing details from which to compare the coupon email validation.

I agree that the billing email is not entered at this point. I am not sure if there is a way around it. The earliest we can do the validation is when the Paypal modal closes and we reach this page:https://d.pr/i/NlJzjz 

At this point, before the "Place Order" can be clicked, the validation could happen.

In fact, while testing I found that one can add a non-owned coupon at this point.

Your thoughts?

chickenn00dle:
Great point @menakas!

I added an additional validation check upon coming to the page you pointed out: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/705/commits/f8cc36dfc4f39ff97b8b1aaa32b89c06f47a67c6

> In fact, while testing I found that one can add a non-owned coupon at this point.

As for validating additional non-owned coupons as soon as they are entered at this stage, I'm not sure this is in the scope of PPEC. It looks like regardless of the payment gateway, a user can enter a non-owned coupon from the checkout page prior to finalizing a payment. 

Perhaps we can open an issue in Woo for this specific problem?

FritzHerbers:
@chickenn00dle 
@menakas 

> Perhaps we can open an issue in Woo for this specific problem?

From the reply received in another issue from WooCommerce, I don't know if there is some kind of hook to set for a recalc/re-validation, as at the end it got unclear:
https://github.com/woocommerce/woocommerce/issues/25777

But in the plugin having that issue, they fixed it:
https://wordpress.org/support/topic/fees-order-not-recalculated-on-later-payment-via-other-gateway/

Maybe this helps to get a step further and having a re-validation before going to PayPal.

chickenn00dle:
Just realizing this is still open. Gonna close as its been quite some time since I've worked on this, or Woo for that matter.

PR #707, opened 2020-04-03
Allow WC to perform validation when missing fields

### LINKED ISSUE #515: When required addon is not selected for a product, mimic add to cart button behaviour
When an addon field is required for a product (using Product Addons Extension), when trying to click 'Add to Cart' a popup shows up explaining that the addon field must be selected.

Here's a gif showing that:

product-add-on-required-before-adding-to-cart

The PayPal buttons are disabled. Instead they should mimic this behaviour.

Initially reported in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/507#issue-233410175

chickenn00dle:
Per #515, when attempting to pay with PayPal from product page where products are missing required or variable fields, the PPEC buttons don't trigger the same behaviors as the default WC Add To Cart button. This leads to some disconnect between the two, and a somewhat awkward experience for the user.

### Steps to reproduce:

1. Create a new required field (Multiple Choice/Radio Buttons) for all products with WooCommerce Product Add-ons extension
2. From the front-end, visit a simple product page and select either of the PayPal buttons. Nothing will happen
3. Select the Add to cart button. You should see an HTML form validation popup
4. Visit a variable product page, and select either PP button (without modifying the product form). Again, nothing will happen
5. Select the Add to cart button. You should see some sort of warning about selecting product options
6. Select product options, leaving the required field empty, and select a PP button again. Nothing should happen.
7. Select Add to cart. You should see another validation warning

### Description

This adds an event listener to the PPEC button container when it is disabled (not all required/variable fields are complete) which triggers the Add To Cart button to click when a PP button is selected. This, in turn, enforces the default WC behavior, displaying the same alerts and popups.

Closes #515

### DIFF
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -27,7 +27,7 @@
 		.on( 'disable', function() {
 			$( '#woo_pp_ec_button_product' )
 				.css( {
-					'cursor': ( variation_valid ? 'pointer' : 'not-allowed' ),
+					'cursor': 'not-allowed',
 					'-webkit-filter': 'grayscale( 100% )', // Safari 6.0 - 9.0
 					'filter': 'grayscale( 100% )',
 				} )
--- a/assets/js/wc-gateway-ppec-generate-cart.js
+++ b/assets/js/wc-gateway-ppec-generate-cart.js
@@ -14,20 +14,29 @@
 
 	$( '#woo_pp_ec_button_product' )
 		.on( 'enable', function() {
-			$( '#woo_pp_ec_button_product' ).css( {
-				'cursor': '',
-				'-webkit-filter': '', // Safari 6.0 - 9.0
-				'filter': '',
-			} );
-			$( '#woo_pp_ec_button_product > *' ).css( 'pointer-events', '' );
+			$( '#woo_pp_ec_button_product' )
+				.css( {
+					'cursor': '',
+					'-webkit-filter': '', // Safari 6.0 - 9.0
+					'filter': '',
+				} )
+				.off( 'mouseup' )
+				.find( '> *' )
+				.css( 'pointer-events', '' );
 		} )
 		.on( 'disable', function() {
-			$( '#woo_pp_ec_button_product' ).css( {
-				'cursor': 'not-allowed',
-				'-webkit-filter': 'grayscale( 100% )', // Safari 6.0 - 9.0
-				'filter': 'grayscale( 100% )',
-			} );
-			$( '#woo_pp_ec_button_product > *' ).css( 'pointer-events', 'none' );
+			$( '#woo_pp_ec_button_product' )
+				.css( {
+					'cursor': ( variation_valid ? 'pointer' : 'not-allowed' ),
+					'-webkit-filter': 'grayscale( 100% )', // Safari 6.0 - 9.0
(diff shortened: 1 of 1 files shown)


assets/js/wc-gateway-ppec-generate-cart.js
```diff
@@ -14,20 +14,29 @@
...
 		.on( 'disable', function() {
-			$( '#woo_pp_ec_button_product' ).css( {
-				'cursor': 'not-allowed',
-				'-webkit-filter': 'grayscale( 100% )', // Safari 6.0 - 9.0
-				'filter': 'grayscale( 100% )',
-			} );
-			$( '#woo_pp_ec_button_product > *' ).css( 'pointer-events', 'none' );
+			$( '#woo_pp_ec_button_product' )
+				.css( {
+					'cursor': ( variation_valid ? 'pointer' : 'not-allowed' ),
```
mattallan, line 30: Hey @chickenn00dle this PR is awesome 💯 
I tested this on variable products and it's looking much better now. Woo!

One question I had was with this `variation_valid` condition being used to display either a pointer or the not-allowed cursor.

What was the reasoning for this?

When I was testing other 3rd party products like a Mix and Match product, I noticed the cursor has changed going from master to this branch. Here's a video starting on master: https://d.pr/v/ICC8tc

It's not a big issue but I'm curious to understand why this was needed and if I'm missing something 🤔 ? Was there a need for showing a pointer cursor when the `variation_valid` is set to true but the button is still disabled?

chickenn00dle: Thanks @mattallan!

> One question I had was with this variation_valid condition being used to display either a pointer or the not-allowed cursor.

I added this condition to mimic the Add to cart button in cases where product addons were being used:

Screen Capture on 2020-05-07 at 10-14-35

I only tested with this extension as it was the one brought up in the original issue. That said, I'll see if I can find an alternative that works more universally.

chickenn00dle: Hmm. It looks like Mix and Match Products is a paid extension, so I'm not able to test on my end. But my immediate thoughts are to default to `not-allowed` here unless both `variation_valid` and `fields_valid` are true:

`'cursor': ( fields_valid && variation_valid ? 'pointer' : 'not-allowed' ),`
 
This would result in a pointer mismatch in the case of Product Addons but would still make sense for the user. It would also remove the inconsistent pointer mismatch you pointed out with Mix and Match Products (I'm assuming).

What are your thoughts?

mattallan: Thanks for sharing this @chickenn00dle I wasn't aware there was a different in pointers for Mix and Match and Product Add-ons.

This is a pretty minor change overall but I think your latest approach sounds good to me:

```
'cursor': ( fields_valid && variation_valid ? 'pointer' : 'not-allowed' ),
```

chickenn00dle: Great! 

I made changes in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/707/commits/909b3f04bed5421de69be9d86acdd428808ac228

I realized if both `fields_value` and `variation_valid` are true, the buttons will be enabled. So I removed the conditional check altogether.

mattallan:
Thanks @chickenn00dle ! This LGTM 💯

PR #717, opened 2020-04-21
Notify user on single product page if cart has contents

### LINKED ISSUE #498: Single product Buy Now button clears the cart of other products
This may be intended behavior given how PayPal Express Checkout works, but I'd like to get dev input on it, as well as possible changes for a better customer experience.

Here's the flow:

1. Add a product to the cart
2. Go to a product and choose to Buy Now
3. Go through PayPal authentication
4. Note that the cart only contains the product you chose to Buy Now and not the previously added product

This is confusing to customers who would be wondering where the other products went, as well as shop owners who would then don't have any way to point out to the customer that the other product(s) were removed. 

Would it be possible to restore the original cart in some way? Or show customers which products were removed?

menakas:
**Issue**: #498

---

### Description

When there exists items in the cart, clicking the smart payment button on single product page cause the cart to be emptied without any warning and the single product is checked out.

This work adds a confirmation message asking if cart contents should be added to the purchase. Based on the user input, either the cart contents are added to the purchase or just the single product is checked out.

### Steps to test:

1. Add a product A to the cart
1. Go to another product B's page.
1. Click on the PayPal smart button.
1. On master, the product A is removed from the cart and B is checked out. On this branch, a confirmation message is asked. If user input is 'Ok' to add cart contents, both A and B are checked out. If user input is 'Cancel', only B is checked out.

### Documentation

Closes #498 .

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -31,6 +31,7 @@ class WC_Gateway_PPEC_Cart_Handler {
 		add_action( 'widget_title', array( $this, 'maybe_enqueue_checkout_js' ), 10, 3 );
 
 		if ( 'yes' === wc_gateway_ppec()->settings->checkout_on_single_product_enabled ) {
+			add_action( 'woocommerce_before_main_content', array( $this, 'display_notice_product' ) );
 			add_action( 'woocommerce_after_add_to_cart_form', array( $this, 'display_paypal_button_product' ), 1 );
 			add_action( 'wc_ajax_wc_ppec_generate_cart', array( $this, 'wc_ajax_generate_cart' ) );
 			add_action( 'wp', array( $this, 'ensure_session' ) ); // Ensure there is a customer session so that nonce is not invalidated by new session created on AJAX POST request.
@@ -278,6 +279,17 @@ class WC_Gateway_PPEC_Cart_Handler {
 			$customer->set_billing_postcode( $billing_postcode );
 			$customer->set_billing_phone( $billing_phone );
 			$customer->set_billing_email( $billing_email );
+		}
+	}
+
+	/**
+	 * Display notice on single product page if cart contains items
+	 *
+	 * @since 2.0.0
+	 */
+	public function display_notice_product() {
+		if (WC()->cart->get_cart_contents_count() > 0) {
+			wc_add_notice( apply_filters( 'woocommerce_paypal_express_checkout__cart_not_empty_message', esc_html__('Clicking on the Smart Payment buttons will not add the cart items for purchase!', 'woocommerce-gateway-paypal-express-checkout' ) ), 'notice' );
 		}
 	}
 


chickenn00dle:
I tested this on my end and ran into the following issue:

modal_cart_contents

It looks like the PP modal is opening at the same time as the browser rendered confirm popup. As a result, before I can confirm, I have to move the modal. I should note, at first, I completely missed the split second where the confirm popup appeared before the modal hid it and thought the script failed or something.

One other small note about the confirm popup:

confirm_cart_contents

This may be a bit nitpicky, but I think from a user's point of view, cancel typically indicates they will be backed out of the process altogether. So the above can be a bit confusing.

If we do want to provide the user with some options, maybe we could use something like jquery's dialog functionality to customize these buttons. 

Or alternatively, maybe we can instead default to emptying the cart as we already are, and instead, provide a notice to the user somewhere letting them know checking out from the product page will empty their cart?

menakas:
@chickenn00dle 
Thanks for your inputs.
>This may be a bit nitpicky, but I think from a user's point of view, cancel typically indicates they will be backed out of the process altogether. So the above can be a bit confusing.

I agree completely.

I've added a commit that displays a notice on a single product page if the cart is not empty. Let me know what you think.

chickenn00dle:
Looking good. 

About the notice text, we may want to specify we are referring to PayPal as not all users will know what `Smart Payment Buttons` are. 

Additionally, perhaps we can be a bit more clear about what will happen once the user selects a button:

- Their cart will be emptied
- They will begin the payment process for the currently viewed product

I'm not sure how to word this exactly, but maybe something along the lines of:

"Selecting a PayPal Smart Payment button will empty your cart and checkout [product name]"

What do you think?

mattallan:
I'm a little unsure about displaying a notice on every product page when the cart contains items.
Some issues:
 - emptying the cart via the mini-cart still shows the notice on the page
 - if you're viewing the same product page that already exists in your cart, it doesn't really make sense to show this notice
 - similarly to the above point, when you have disabled the "Redirect to the cart page after successful addition" setting and click add to cart, the page refreshes with this notice 🤔 

I can see customers being confused by this, but also some store managers not really liking this behaviour either.

I actually liked the first idea on this PR with displaying an alert when a customer clicks the SPB when they already have cart items. What happened to that? Was there no possible workaround for this issue: 



As for wording on the alert, we could say something like:

"You are about to clear your cart and purchase [qty] x [product name] with PayPal. Do you wish to continue?"

---

Something else that was brought up on support forum was displaying some sort of text or heading just above Smart Payment buttons like:

"Purchase this product with PayPal"

menakas:
>I actually liked the first idea on this PR with displaying an alert when a customer clicks the SPB when they already have cart items. What happened to that?

@mattallan 
Me too. I liked the idea. But I had a roadblock on it. From what I gather from reading through PayPal documentation, when the SPB is clicked, the validation is already done and the payment flow is initiated. So, doing validation on clicking the button is not straightforward to implement.

The other option for synchronous validation suggested is to disable the button and enable it only when the validation is done.

Now I am confused. There are 3 options. Which is the best way forward?

jorgeatorres:
@menakas: I'm not entirely sure on how to do this in the current checkout.js implementation (soon to become the "legacy" implementation), but with the [new JS SDK](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/668) this should be possible (and easier). I believe adding an `onClick` handler along these lines should work:

```js
onClick: function( data, actions ) {
	if ( ! window.confirm( "<Message here>. Do you wish to continue? ") ) {
		return actions.reject();
	}

	return actions.resolve();
}
```

Maybe it might be best to wait until #668 is merged before revisiting this issue. Thoughts @mattallan?

mattallan:
Nice @jorgeatorres and thanks for chiming in on this 😄 

> Maybe it'd be best to wait until #668 is merged before revisiting this PR. Thoughts @mattallan?

Your approach above sounds good and yeah, I would agree with waiting until #668 is merged before revisiting this :)

@menaka on top of Jorge's code above you can find more docs on using onClick handler [here](https://developer.paypal.com/docs/checkout/integration-features/validation/#) :)

jorgeatorres:
Hi everyone!

I did some research on this and turns out it's not really possible to confirm the intention via a confirm() prompt.
According [to the docs](https://developer.paypal.com/docs/checkout/integration-features/validation/#synchronous-validation), there are two ways to implement validation: sync and async. Now...

- To do sync validation, the logic has to be implemented inside `onInit()`, which means we need to know whether the user wants to clean their cart *before* they click the PayPal button. That means we would have to show a checkbox or something close to the PayPal button that they'd have to click and then they'd be able to click the PayPal button. This, IMHO, makes no sense.
- When doing async validation, it is true that you're not taken to paypal.com until the promise has been resolved (i.e. the user clicks "OK" in a Promise-wrapped prompt() dialog). This sounds great, but turns out (1) the PayPal window remains open the whole time with a "loading spinner" until you've either clicked OK or Cancel (2) the PayPal window is focused, so the prompt() is obscured by the window itself 🤷. This is not in error, but by design:

   > Whenever possible, PayPal recommends setting up synchronous validation for your page. This results in the best possible user experience because the buyer sees instant feedback and no popup window displays.

   > Do not use asynchronous validation for any validation that can be done synchronously as it degrades the buyer's experience.

   You can see this at work in these two screencasts (one for clicking 'Cancel' and one for clicking 'OK'):

   screencast2

   screencast3

As such, I believe the only alternative right now would be to display a message next to the button, which in line with the discussion above, is not the ideal UI/UX. Because of this, I'm closing this PR. If we eventually decide to go with the "message" option or PayPal changes how this aspect of the SDK works, a new PR can be submitted.

PR #719, opened 2020-04-23
Update phpcs config and add Travis integration

jorgeatorres:
### Description

This PR updates the PHP Code Sniffer config to use the latest WooCommerce standards and also integrates with Travis so that `phpcs` is ran against `master` and `2.0_dev`.

The Travis integration involves two jobs:
- A job runs `phpcs` on the range of commits that were included in the push or pull request. _This job is not allowed to fail_. The reasoning behind this is that we should take care of all the smells in files we're touching in a PR, even if they belong to parts we're not modifying.
- Another job runs `phpcs` against the whole repo. This job is allowed to fail and its only purpose is to provide a report of standards compliance for the whole repo, without forcing anyone to address those issues immediately.

### Steps to test:

These PR doesn't touch any core files. As such, plugin functionality should remain intact.
For development, it is possible to test that Composer support is working by running `composer install` in the repo directory. Tests can be run via `composer phpcs`.


mattallan:
> The Travis integration involves two jobs

I didn't know this was possible but this is awesome 😻
Thanks @jorgeatorres for getting this all setup across our new repos!

PR #720, opened 2020-04-23
[Coding Standards] Fix smells

jorgeatorres:
# Description

This PR fixes some of the WC standards violations in our codebase.
In particular, it addresses smells with fixes that I thought were easy to review in batches or didn't require important changes to the codebase.

Even though this PR doesn't make serious changes to the codebase it has a lot of commits and touches lots of lines. As such, if reviewing this PR proves to be too difficult I'm ok with us dropping it and relying entirely on the approach suggested in #719. If someone does feel like reviewing it, I should say that merging this would make everyone's job easier in the near future as this PR reduces the number of violations from ~1100 to ~600.

**Tip:** Reviewing by commit instead of by file might be easier.


jorgeatorres:
@mattallan: I've rebased this on the latest master and fixed the conflicts. I also fixed other smells, so we should now be 100% standards compliant.

> Tip: Reviewing by commit instead of by file might be easier.

I'm not sure if this is still true 😅, however.

mattallan:
Thanks @jorgeatorres for working on these changes and for keeping this PR up to date! :100:

I also greatly appreciate that you used `// phpcs:ignore/disable` to reduce the possibility of introducing a breaking change.

I went over each line that was changed in this PR but for my own sanity I did a general test on this branch:
 - settings page loading and saving properly
 - check hiding funding methods and different button settings
 - paying with PayPal and credit card from checkout page
 - paying with PayPal and credit card from product page
-  refund orders
 - authorise and capture
 - subscription purchases
 - free orders

All was good :+1: merging

PR #730, opened 2020-05-12
Remove the Client ID and Secret settings and related code

mattallan:
**This PR is still a Work In Progress.** 

---

# Description

Now that we've merged the new SDK upgrade changes, one of the final things leftover on that PR was to remove the Rest Client ID and Secret settings and all related code, this PR does that.

**Changes**

41d2efe - Removes client ID and secret settings and any references found in the `ppec-settings.js`

60cd29d - Removes the `WC_Gateway_PPEC_REST_Client_Credential` and `WC_Gateway_PPEC_REST_Client` classes. Since we're no longer storing and validating the Client ID and Secret, we don't need these classes.

593f986 - Removes `$settings->get_sandbox_rest_api_credentials()` and updates other similar functions on the settings class so they can be used to fetch the live Client ID.

As part of this commit, I noticed `$use_js_sdk` inside the `enqueue_script()` function was never  used so I removed that and directly set the `client-id` script arg to:

```
'client-id'  => $settings->get_active_rest_client_id(),
```

# Todos

- [x] set the Woo Client ID in `get_live_rest_client_id()`


mattallan:
@jorgeatorres I have updated the Client ID so this is ready for review and testing now 😃 

I also received the following Woo PPEC sandbox client ID from Paul:

```
AauZt9rTsAWHqG4l6NkXbRVHfAYGbZpBDpIqxNBoyuD2gDvlH-aPY48C70_s2eSzwndHpQnuFjS-mRjt
```

For testing purposes, I used this Woo client ID instead of `sb` and the checkout worked fine.
I also double-checked the transaction looked normal in my merchant account :+1:

jorgeatorres:
Hi @mattallan!

This looks good 💯. Given we're actually generating the charges and everything via NVP on the backend, I think using just our Client ID for the JS script should be ok, but I was wondering whether we should also pass the [merchant-id parameter](https://developer.paypal.com/docs/checkout/reference/customize-sdk/#merchant-id) just in case.

If PayPal Checkout is correctly configured, we should be able to get the Merchant ID by calling `get_payer_id()` on an instance of `WC_Gateway_PPEC_Client`. [One is readily available](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/8c785be01da8b3658b3850aede23e44196c42768/includes/class-wc-gateway-ppec-cart-handler.php#L471) close to where we enqueue the script.

What are your thoughts on this?

mattallan:
Thanks @jorgeatorres for raising this!!
Based on the infor in that link you shared above, I think we should definitely be sending the merchant ID.

Nice catch 💯 

In the table below, we fit under the partner/marketplace integration which states we **must** pass the merchant ID.


I've pushed up these changes in b983eaf and tested the checkout again  ✅

PR #731, opened 2020-05-19
Merging 2.0 into master

mattallan:
I noticed some of the commits have been duplicated:

- a6d2970, b8cc51c - Check if sku value is set before adding it as a param
- a2b8eda, 476baca - Fix whitespace
- 1551feb, 515fd3c - Add button label as a setting
- 8ad2013, 0cbaa17 - Display label in horizontal and vertical layout
- 38c868d, 3c145c8 - Pass the button label to be displayed on the smart payment buttons
- 3d9ae72, d08bc72 - Check if product_page shortcode is present to determine if it is a product..
- 171e5ad, 8c57376 - Use wc_post_content_has_shortcode() to check for shortcode
- 1af4ec5, 566dba3 - Pass the product SKUs to PayPal requests

I'm not sure how this might have happened but likely caused by a rebase or something when we moved commits from in `1.7_dev` to go into `1.6.21` and then we created a `2.0_dev`. 

Regardless, we should clean up the history and make sure there are no missing PRs/commits.

- [x] remove duplicate commits from history
- [ ] double-check there are no missing PRs/commits in `2.0_dev`


james-allan:
How did you find those duplicates @mattallan? Any ideas on how we might be able to check if commits are missing? Tagging @menakas in case she has some ideas using regex or terminal commands.

mattallan:
> How did you find those duplicates @mattallan?

These are just the duplicates commits I noticed in this PR.
I found them by scrolling through the commits and noticed 2 x `Add button label as a setting`. I found the rest by just manually sifting through the commits above

> Any ideas on how we might be able to check if commits are missing?

I was just going to go with a very manual process, which is to look at each PR that went into 2.0 or 1.7 and make sure it exists in master or this PR 😅

menakas:
>How did you find those duplicates @mattallan? Any ideas on how we might be able to check if commits are missing? Tagging @menakas in case she has some ideas using regex or terminal commands.

@james-allan Yes. it can be done. 

Say, we want to find the duplicates among the last 200 commits based on commit message.
Assuming we don't have aliases for log and it prints one line per commit , with the commit id followed by the commit message,
`git log -n200 | sort -k2 | uniq -d -f1`

`git log -n200` to get the last 200 commits oneline. 
`sort -k2` sort the lines based on field2 (the commit message).
`uniq -d -f1` -d flag is to report duplicated lines only, -f1 do the comparisons after ignoring field 1(do not use commit id for comparison)

This gives a list of the duplicate commits(actually one of the duplicates only).

Screenshot 2020-05-19 at 11 14 39 AM

mattallan:
Ooo thanks @menakas! That's good to know :)

I have rebased this already so there shouldn't be any duplicates now

The culprit was a manual merge I think I did early on which had the commit message `Merge branch 'feature/make-spb-default' into 1.7_dev`. This single merge included other merge commits, which were then merged manually later on 😖 

Thanks @james-allan as well for the extra  👀

menakas:
Here is another option where you can get both the commits.

`git log -n200 | sed 's/ /\t/' | sort -k2 -t'\t' --stable | gawk -F'\t' '{if ( $2==old ) { print $0; print oldcommit,old }; old=$2; oldcommit=$1 }'` on *nix.

On Mac, replace the `\t` representing tab to `Ctrl+V tab`

`git log -n200` to get the last 200 commits in one line format.
`sed 's/ /\t/' ` to replace the first space with tab as delimiter
`sort -k2 -t'\t' --stable` sort the lines based on field2 (the commit message), the --stable flag is to ensure that after sorting duplicate lines appear in the same order as before.
`gawk -F'\t' '{if ( $2==old ) { print $0; print oldcommit,old }; old=$2; oldcommit=$1 }'` get the duplicate lines

Screenshot 2020-05-19 at 2 28 41 PM

PR #740, opened 2020-05-25
Fixes `paypal.getFundingSources is not a function` console errors

### LINKED ISSUE #739: Console error: Uncaught TypeError: paypal.getFundingSources is not a function
### Describe the bug

Customers have started reporting a javascript error after updating to 2.0.0. See first report: https://wordpress.org/support/topic/latest-version-has-js-critical-error/

I have been able to replicate this by activating PayPal Powered by Braintree Gateway which is still using the older `checkout.js` file 😢 

Steps to replicate:
1. Activate PayPal Powered by Braintree gateway and set it up in sandbox mode
   1. _If you don't have PayPal Braintree you should be able to just update [this condition](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/d958f601a32f5a2cdf6319186b8b7205c518a96b/includes/class-wc-gateway-ppec-cart-handler.php#L535) to always fail by adding `false &&`._
2. Enable smart payment buttons on the checkout page
3. Set **Hide payment methods** setting to have 1 or more payment options disabled: https://d.pr/i/xuUmyk
4. Load the checkout page 

You should notice that no PayPal buttons are loaded and the error in console:

mattallan:
**Issue**: #739 
**Ticket**: https://wordpress.org/support/topic/latest-version-has-js-critical-error/

---

### Description

Customers have reported that after upgrading to 2.0.0 their PayPal buttons aren't loading and they're seeing the following console/javascript errors:

```
Uncaught TypeError: paypal.getFundingSources is not a function
    at render (wc-gateway-ppec-smart-payment-buttons.js:208)
    at HTMLBodyElement.dispatch (jquery.js:3)
    at HTMLBodyElement.r.handle (jquery.js:3)
    at Object.trigger (jquery.js:3)
    at Object.a.event.trigger (jquery-migrate.min.js:2)
    at HTMLBodyElement.<anonymous> (jquery.js:3)
    at Function.each (jquery.js:2)
    at a.fn.init.each (jquery.js:2)
    at a.fn.init.trigger (jquery.js:3)
    at Object.success (checkout.min.js:1)
```

I discovered that this issue is being caused by the old `https://www.paypalobjects.com/api/checkout.js` script being loaded even though we have set `wc_ppec_context.use_checkout_js` to `false` (meaning we're using sdk functions on the `paypal` variable).

At first I thought this was a caching issue, but after some investigation, I found that there are other PayPal extensions which are likely loading file. Just doing a search on my local dev site I found [the following plugins](https://d.pr/i/qRT1vB) that still use/load this incompatible `checkout.js` file
 - Jetpack Simple Payments
 - CartFlows Pro
 - PayPal powered by Braintree
 - PayPal Checkout (only when using legacy filter)

In order to have the new Javascript SDK and Checkout.js work on the same page, PayPal recommends the following:

> To load this sdk alongside the existing version (checkout.js), please specify a different namespace in the script tag, e.g. <script src="https://www.paypal.com/sdk/js?client-id=CLIENT_ID" data-namespace="paypal_sdk"></script>, then use the paypal_sdk namespace in place of paypal in your code.

This PR fixes the issues by using the `paypal_sdk` namespace approach as suggested.

### Steps to test:

1. Activate PayPal Powered by Braintree gateway and set it up in sandbox mode
2. Enable smart payment buttons on the checkout page
3. Set Hide payment methods setting to have 1 or more payment options disabled: https://d.pr/i/xuUmyk
4. Load the checkout page

**On master**

**On `isssue_739`**

Closes #739.

### DIFF
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -2,6 +2,9 @@
 ;( function ( $, window, document ) {
 	'use strict';
 
+	// Use global 'paypal' object or namespaced 'paypal_sdk' as PayPal API (depends on legacy/SDK mode).
+	var paypal = wc_ppec_context.use_checkout_js ? window.paypal : window.paypal_sdk;
+
 	// Show error notice at top of checkout form, or else within button container
 	var showError = function( errorMessage, selector ) {
 		var $container = $( '.woocommerce-notices-wrapper, form.checkout' );
@@ -43,7 +46,7 @@
 
 		var paypal_funding_methods = [];
 		for ( var i = 0; i < methods.length; i++ ) {
-			var method = ( wc_ppec_context.use_checkout_js ? paypal : paypal_sdk ).FUNDING[ methods[ i ].toUpperCase() ];
+			var method = paypal.FUNDING[ methods[ i ].toUpperCase() ];
 			if ( method ) {
 				paypal_funding_methods.push( method );
 			}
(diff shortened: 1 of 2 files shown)


jorgeatorres:
👋 @mattallan:

This looks good to me and works correctly 💯.
I pushed a small commit to simplify the code inside the JS file, so that we use the namespaced PayPal object (`paypal_sdk`) but can refer to it simply as `paypal`, at least inside our functions. Since it is a local variable, it shouldn't make any difference in terms of functionality, but reduces the changes a bit.

I'm off today, so I'm just going to approve the PR for now. With or without my commit, I think this is ready to go. Feel free to either keep it/remove it tomorrow when you get to this and do the actual merge.

mattallan:
Thanks very much @jorgeatorres for the review and commit, I've tested the latest changes and confirmed the fix again.

LGTM

PR #758, opened 2020-05-29
Don't display the Debit or Credit card buttons when button layout setting is Horizontal

mattallan:
### Description

While hanging out in WP.org support I've been noticing some complaints about the Debit or Credit cards buttons being displayed after updating to 2.0:
 - https://wordpress.org/support/topic/remove-black-debit-or-credit-card-button-in-paypal-checkout/
 - https://wordpress.org/support/topic/paypal-checkout-2-0-1-new-button-issue/#post-12906795

After comparing 1.6.21 and 2.0.n, I noticed that we started displaying the Debit or Credit card button on the horizontal layout.

Because there's no setting for removing the debit/credit card button on horizontal view and this was an undocumented change, I think we should just restore the previous behaviour from 1.6.21 and hide the card button.

Something interesting I found is that the Paypal demo site doesn't show a Debit card button when it's set to horizontal: https://developer.paypal.com/demo/checkout/#/pattern/horizontal 🤔  
And I've confirmed we're definitely setting `layout` to `horizontal`

### Steps to test:

1. Go to PayPal Checkout settings
1. Set Button Layout to "horizontal"
1. View product page
   1. 1.6.21/legacy: https://d.pr/i/mkBNQY
   1. master: https://d.pr/i/3yZUCR

### Documentation

### DIFF
--- a/includes/class-wc-gateway-ppec-cart-handler.php
+++ b/includes/class-wc-gateway-ppec-cart-handler.php
@@ -463,7 +463,7 @@ class WC_Gateway_PPEC_Cart_Handler {
 			$data['disallowed_methods'] = $data['hide_funding_methods'];
 		} else {
 			$data['allowed_methods']    = 'yes' === $data['credit_enabled'] ? array( 'CREDIT' ) : array();
-			$data['disallowed_methods'] = 'yes' !== $data['credit_enabled'] ? array( 'CREDIT' ) : array();
+			$data['disallowed_methods'] = 'yes' !== $data['credit_enabled'] ? array( 'CARD', 'CREDIT' ) : array( 'CARD' );
 		}
 		unset( $data['hide_funding_methods'], $data['credit_enabled'] );
 


jorgeatorres:
Hi @mattallan!

I looked into this and found things are a bit more complicated. I apologize for the long post.

#### Context

Under normal circumstances "Hide funding method(s)" is only available when you use the **vertical** layout. This layout is also required for the [Credit/Debit card buttons to become available](https://developer.paypal.com/docs/checkout/integration-features/standard-card-fields/):

> Standard Cards Fields display in a vertical Smart Payment Buttons integration.

The horizontal layout, in contrast, should look exactly as in the PayPal demo you linked above: that is, only the PayPal button and -possibly- the PayPal Credit button should be shown.

#### Why is the credit card button appearing here when the "Horizontal" layout has been set?
That's because "Enable PayPal Credit to eligible customers" is not checked. When that happens, we try to hide this button regardless of the layout, but PayPal's JS API is not flexible enough to do [standalone buttons](https://developer.paypal.com/docs/checkout/integration-features/standalone-buttons/) **and** the horizontal layout at the same time: 'style' attributes are ignored and buttons get rendered as in the vertical layout (even though the CSS classes still say "horizontal" 🤷). Credit cards are also "automatically" enabled since they are a valid funding source and we're not strictly in the "horizontal" layout anymore.

#### What do we do now?
The fix you've proposed clearly addresses the problem of hiding the credit card button when using the horizontal layout (and I wouldn't oppose merging as is), but doesn't address the issue of our horizontal layout not being "the" horizontal layout (unless PayPal Credit is allowed). This difference can be relevant when on mobile, where presenting things using the horizontal layout can be important (for instance, for Venmo).

One way to fix this issue would be to globally disable PayPal Credit (when needed) using the `disable-funding` script argument. This would work for both horizontal and vertical layouts, makes our horizontal layout match PayPal's and also doesn't prevent context-specific funding methods to be hidden when the vertical layout is in use in those contexts.

The only problem I see with this approach is that we also have an "Enable PayPal Credit to eligible customers" checkbox in the context-specific settings for the mini-cart, checkout, etc. and as such we can't use the `disable-funding` argument unless we make the PayPal Credit setting a global one (and one that applies to both horizontal and vertical layouts). We lose a bit of flexibility there but it might be a good compromise: I don't know how common is for merchants to want to use the horizontal layout for different contexts and at the same time hide PayPal Credit in one but not the other. And even if it is, they could just do the vertical layout in those cases.

What are your thoughts?

jorgeatorres:
An addendum to the above:

As I suspected in #774, I was able to confirm that the "layout" style option does have some influence on the methods that are available by default. For instance, after you apply #774, the debit/credit card button is no longer displayed when the button layout setting is "horizontal". That is, without making any other changes to the code.

So maybe we can merge #774 (or a variation of it) instead.

That said, there's still something amiss (even with #774): when we use the default horizontal layout configuration without hiding any methods, the SDK automatically decides which methods to show and creates the 1 or 2 columns needed, but when using standalone buttons this doesn't happen and the buttons are apparently rendered vertically. This is something we can workaround in our code if necessary, but that we wouldn't have to do with what I proposed in my previous comment. What do you think is best?

mattallan:
Wow, thanks @jorgeatorres for this detailed response. I had a bad feeling there was something else happening here...

> I don't know how common is for merchants to want to use the horizontal layout for different contexts and at the same time hide PayPal Credit in one but not the other

Yeah I agree. I think it would be pretty rare for this to be the case, but I wouldn't put it past some merchants making the most out of the flexibility we have to offer in our settings 😄 

I'm a little concerned with making a change like moving the enable PP Credit in horizontal layout to apply to all contexts, especially in a minor patch. So I'm all for finding alternative fixes  :)

I was thinking about what the changes look like if we were to use a global context setting for this and here's what it would involve.

 - Remove context-specific "enable PP Credit for eligible customers" setting
 - Add a global setting "enable PP Credit for eligible customers when using horizontal layout"
 - On upgrade, map/migrate existing settings to a new global setting i.e.
   - enable the global setting if there were any "enable PP Credit for eligible customers" checked in the previous version
   - check if the customer had some horizontal layouts contexts with PP Credit enabled and not others (the rare case we talk about above) and then automatically switch them to vertical layout with every other funding method hidden, or just add a notice to recommend they take these steps to preserve their old behaviour 🤔 

IMO that seems a bit messy/complicated.

> What do you think is best?

Yeah, I think we can close this PR in favour of the fixes you added in #774.

There were still some broken cases in this PR, but from my testing, everything was fixed in #774 :100: see below.

Testing setup 
 - I'm setting the `buyer-country` to `DE` and `US` for testing purposes
 - Layout set to horizontal

.

- `master` branch
  - PP Credit disabled/DE customer: https://d.pr/i/6h2G9l :x:
  - PP Credit disabled/US customer: https://d.pr/i/NqPIOe :x:
  - PP Credit enabled (DE customer): https://d.pr/i/6h2G9l ✅ 
  - PP Credit enabled (US customer): https://d.pr/i/3cNI8j ✅ 
- This PR (`card_buttons_horizontal_layout`)
  - PP Credit disabled (DE customer): https://d.pr/i/g1nh8C :x:
  - PP Credit disabled (US customer): https://d.pr/i/Wc4H7T :x:
  - PP Credit enabled (DE customer): https://d.pr/i/U1DJZt ✅ 
  - PP Credit enabled (US customer): https://d.pr/i/4H9Jsc ✅ 
- #774 (`preserve-layout-configuration`)
  - PP Credit disabled (DE customer): https://d.pr/i/0Unh51 ✅ 
  - PP Credit disabled (US customer): https://d.pr/i/UFrs4L ✅ 
  - PP Credit enabled (DE customer): https://d.pr/i/Z0740t ✅ 
  - PP Credit enabled (US customer): https://d.pr/i/ivm0Fk ✅ 

> when we use the default horizontal layout configuration without hiding any methods, the SDK automatically decides which methods to show and creates the 1 or 2 columns needed, but when using standalone buttons this doesn't happen and the buttons are apparently always rendered vertically

Hmm, I might be misunderstanding this but I didn't see this while I was testing (can confirm horizontal columns were still being used). Is this noticeable when loading other funding methods like venmo?

Given the test results, I'm going to close this and merge #774 
Thanks, @jorge!

Derya00:
People want to use the horizontal layout for the Paypal button and at the same time hide PayPal Credit when there is a bug on Vertical layout where the Paypal button does not display (display set to none). This actual issue seemed to have been lost as tickets closed/open. But I suppose the layout to horizontal was not a proper fix in the first place.

jorgeatorres:
Hi @Derya00!

This PR (and related issue) was about the Debit/credit card button appearing when the layout was set to horizontal, which is unlike previous versions. We were not trying to modify the behavior in regards to PayPal Credit, which can still be disabled via a checkbox (even in horizontal mode).

> there is a bug on Vertical layout where the Paypal button does not display (display set to none)

This is the first time I hear about this. Setting the display to none is also not how we "hide" buttons. What we do is not render them at all. This might be the result of some sort of [plugin or theme conflict](https://docs.woocommerce.com/document/how-to-test-for-conflicts/). If not and this is still happening to you, feel free to [open a new issue](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/new) and we will take a look.

PR #761, opened 2020-06-03
Fix currency number format issue

JMMJagers:
### Description

I changed the currency format set in the send info. We are using this plugin for 1 of our sites, but when we tried to implement it for out other site we came across the problem of the currency returning paypal errors about the currency format.
Prices ending in 0 cents got cut off (example: 90.20 -> 90.2) which returned errors due to not having 2 decimals.
Tested the changes myself on both sites. The already working site is still functioning accordingly. The site we had problems with also works after this change.

### Steps to test:
1. Use different Woocommerce settings for decimals. (0, 1 (to make sure a 0 would get cut off) and 2)
2. Use sandbox to confirm a payment.

### Documentation

no

customers wouldn't notice any changes.

Closes # .

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -300,12 +300,12 @@ class WC_Gateway_PPEC_Client {
 		$params = array_merge(
 			$params,
 			array(
-				'PAYMENTREQUEST_0_AMT'         => $details['order_total'],
-				'PAYMENTREQUEST_0_ITEMAMT'     => $details['total_item_amount'],
-				'PAYMENTREQUEST_0_SHIPPINGAMT' => $details['shipping'],
-				'PAYMENTREQUEST_0_TAXAMT'      => $details['order_tax'],
-				'PAYMENTREQUEST_0_SHIPDISCAMT' => $details['ship_discount_amount'],
-				'NOSHIPPING'                   => WC_Gateway_PPEC_Plugin::needs_shipping() ? 0 : 1,
+				'PAYMENTREQUEST_0_AMT'          => set_currency_format( $details['order_total'] ),
+				'PAYMENTREQUEST_0_ITEMAMT'      => set_currency_format( $details['total_item_amount'] ),
+				'PAYMENTREQUEST_0_SHIPPINGAMT'  => set_currency_format( $details['shipping'] ),
+				'PAYMENTREQUEST_0_TAXAMT'       => set_currency_format( $details['order_tax'] ),
+				'PAYMENTREQUEST_0_SHIPDISCAMT'  => set_currency_format( $details['ship_discount_amount'] ),
+				'NOSHIPPING'                    => WC_Gateway_PPEC_Plugin::needs_shipping() ? 0 : 1,
 			)
 		);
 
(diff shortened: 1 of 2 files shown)


mattallan:
Hi @JMMJagers, 

Thanks for reporting the issue and opening a PR

I'm having trouble reproducing the issue you're describing with currencies and order total formats.

Can you please share with us some more details about your current set up to and how you're replicating the error to help us diagnose the problem?

- What's the error you are seeing? Is it an error message on PayPal, our checkout, or unexpected result? Can you send us a screenshot or give us more details?
- What's the currency of your store set to (we have some logic which forces 0 decimal values when rounding based on currency)
- What is the price of your product displayed in the [Edit Product page](https://d.pr/i/3ju5r3), does it match the number of decimals your store is set to.
- Are there taxes or discounts being applied to the cart?
- Are you clicking on the PayPal button or the Debit Card button?
- Do you have [Debug Logs in your PayPal Checkout settings](https://d.pr/i/c5EwrM) enabled? If so, can you send us a snippet of your PayPal express checkout logs? Specifically, the data we're interested in is `WC_Gateway_PPEC_Client::_request` and `WC_Gateway_PPEC_Client::_process_response` which highlights an error.

Apologies for all the questions but hopefully this should help us source the problem.
Thanks for your time!

JMMJagers:
Hi @mattallan ,

I had several PayPal errors in the checkout, being: 10002, 10426, 10427, 10041 and 10431.
We use "EUR" as currency, so it should force the decimals to 2.
The decimals in our woocommerce settings is also set to 2.
Thousand seperator: '.'
Decimals seperator: ','

Product price was 64,95
with shipping of 3,95
so a total of 68,90
Giving me a debug log with this:
(client_request)
    [PAYMENTREQUEST_0_AMT] => 68,9
    [PAYMENTREQUEST_0_ITEMAMT] => 64,95
    [PAYMENTREQUEST_0_SHIPPINGAMT] => 3,95
price on the productpage itself is 64,95.

We have the tax settings as already included in the price we add.

I was using the PayPal button to test the payment.

mattallan:
Hi @JMMJagers, thanks for passing that information on.

My current suspicion is that there's a plugin or theme conflict causing the issues here because when I set my local site up in the same way as you describe, I'm still not seeing any errors when I purchase a product. 

Here's a screenshot of my logs with some highlighted sections to show prices and payment status: https://d.pr/i/9W6NOA

Notice I still get the $68.9 in the `AMT` log, but no issues.

Product price: https://d.pr/i/0Yp01a
Currency settings: https://d.pr/i/TBzx6G
Checkout: https://d.pr/i/V7SCux

Thanks for sharing the error codes you're getting. Here's what each of those codes mean:
 - `10002` - Security Header Invalid, incorrect API credentials
 - `10426` - Transaction refused because of an invalid argument
 - `10427` - Shipping total is invalid
 - `10041` - (couldn't find this error code)
 - `10431` - Item amount is invalid.

Something I noticed with the values you shared above is the `,` decimal is showing up in your logs which is not something that should happen:

```
[PAYMENTREQUEST_0_AMT] => 68,9
[PAYMENTREQUEST_0_ITEMAMT] => 64,95
[PAYMENTREQUEST_0_SHIPPINGAMT] => 3,95
```

Based on this, it appears that there's some plugin or theme that may be filtering the returned values from calling `$order->get_subtotal()` and other `get_prop()` functions and applying some formatting to them. 🤔 

Does this sound like a possibility?

If you have a plugin that needs to do this, we can fix this by doing something like you're doing in this PR (i.e. making sure the format is correct), but I suspect there are other issues that could be happening on your site as a result of this.

JMMJagers:
@mattallan 
if this problem gets caused by another plugin we are using I get confused, because it works fine on another website of ours. We use all the same plugins on both.
Never the less I will try and take a look if there are more differences between them, seems like the best explanation as to why it is causing us errors.

PR #764, opened 2020-06-04
save order-data to database

JMMJagers:
### Description
Refunds did not save to database with WC_versions over 3.0 due to missing code.
The method got updated but not saved to the database.

### Steps to test:
1. Place an order with paypal as payment method.
2. Check database table meta-data for _woo_pp_txnData -> refunded_amount (is 0)
3. Do a partial or complete paypal refund
4. Check database table meta-data for _woo_pp_txnData -> refunded_amount

Closes # .

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -414,6 +414,7 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 						update_post_meta( $order_id, '_woo_pp_txnData', $txn_data );
 					} else {
 						$order->update_meta_data( '_woo_pp_txnData', $txn_data );
+						$order->save();
 					}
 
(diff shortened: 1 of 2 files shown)


mattallan:
Wow @JMMJagers 🙀 

It seems this bug has been around for at least 3 years which makes me sad 😭 
Thanks very much for this PR.

I've tested full and partial refunds and all LGTM :+1:

---

Just making a note for others looking at this.

Along with fixing an issue in our `process_refund()`, this PR also adds a `$order->save()` to our `wc_gateway_ppec_save_transaction_data`, but with how we use this function there's not necessarily a bug here because `$order->save()` will eventually be called in subsequent calls to `$order->update_status()` and `$order->payment_complete()` later on (see `do_payment()` and `_process_reference_transaction_response()`)

But since this is a public function found in `functions.php` there may be a third party using this function expecting the order data to be saved.

PR #771, opened 2020-06-15
Change the label and description of the PayPal Mark setting

### LINKED ISSUE #675: Make description of "Enable PayPal Mark" clearer
Via `p1580974407184800-slack-lambda`:

> Hi team :wave:
> I’m a bit confused with PayPal Checkout behaviour and wondering if you can help. So even if PayPal Checkout is enabled as a gateway and properly configured, it won’t show up on checkout until "Enable the PayPal Mark on regular checkout" is checked.
> [Our docs say](https://docs.woocommerce.com/document/paypal-express-checkout/#section-7) that this setting only adds a logo to the checkout page, not enable/disable PayPal Checkout at all:
>
> > 6. Tick the checkbox to Enable PayPal Mark if you would like to display the PayPal name during checkout on your site, which can give customers assurance.
>
> [...]
> If that’s indeed by design, we need to at least update the docs and the ? wording in wp-admin. > [...]
>
>image

----

While on the settings screen the "Enable the PayPal Mark on regular checkout" checkbox is checked by default, maybe we should make it clear that it controls whether PayPal EC appears as an option during checkout or not.

As suggested, we can achieve this by adjusting our docs and/or the on-screen setting description.

rakmob42:
### Description

This fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/675. This is a simple change to the label and description of the "PayPal Mark" setting.

Previous:

> Label: Enable the PayPal Mark on regular checkout.
> Description: This enables the PayPal mark, which can be shown on regular WooCommerce checkout to use PayPal Checkout like a regular WooCommerce gateway.

New:

> Label: Show PayPal Checkout payment option on the regular WooCommerce checkout page.
> Description: Disabling this will hide PayPal mark from the regular WooCommerce checkout page.

I'm not sure whether I should call it `PayPal payment option` or `PayPal mark`. I think the `PayPal payment option` is much simpler to understand.

### Documentation

Currently, there's no mention of "PayPal Mark" in our documentation. We can include it here: https://docs.woocommerce.com/document/paypal-express-checkout

Closes #675

### DIFF
--- a/includes/settings/settings-ppec.php
+++ b/includes/settings/settings-ppec.php
@@ -461,11 +461,10 @@ $per_context_settings['button_size']['class']    .= ' woocommerce_ppec_paypal_sp
 $per_context_settings['credit_enabled']['class'] .= ' woocommerce_ppec_paypal_spb';
 
 $settings['cart_checkout_enabled'] = array(
-	'title'       => __( 'Checkout on cart page', 'woocommerce-gateway-paypal-express-checkout' ),
+	'title'       => __( 'Enable on the cart page', 'woocommerce-gateway-paypal-express-checkout' ),
 	'type'        => 'checkbox',
 	'class'       => 'woocommerce_ppec_paypal_visibility_toggle',
-	'label'       => __( 'Enable PayPal Checkout on the cart page', 'woocommerce-gateway-paypal-express-checkout' ),
-	'description' => __( 'This shows or hides the PayPal Checkout button on the cart page.', 'woocommerce-gateway-paypal-express-checkout' ),
+	'label'       => __( 'Enable PayPal Checkout buttons on the cart page', 'woocommerce-gateway-paypal-express-checkout' ),
 	'desc_tip'    => true,
 	'default'     => 'yes',
 );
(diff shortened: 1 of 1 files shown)


james-allan:
Great first PR @codestor4 💯 

I agree that the previous language was pretty confusing. After looking at all the settings together though, I think we should use this opportunity to make the setting names more consistent. 

Screen Shot 2020-06-17 at 3 54 34 pm

link: https://cloudup.com/cYoyaKbf0QR

All of the highlighted settings enable the PayPal buttons on the respective pages and they are all worded differently.

In my opinion using the simplest wording here is probably be the best approach. `Checkout on the x page` and `Enable PayPal Checkout on the x page` is concise and clear. The only issue with this is when it comes to the checkout settings. `Checkout on the checkout page` 😕 

What do you think about this **Enable on the Checkout Page** and **Enable PayPal Checkout on the regular checkout page.**

That would look like this:

Screen Shot 2020-06-17 at 4 27 14 pm

https://cloudup.com/cpghvgKhKsG

While we're here we should update the tooltip text too. For example https://cloudup.com/c_VUHFanwnB vs https://cloudup.com/cIrwakzLNz7 vs https://cloudup.com/cT6IgwnhjJ4

codestor4:
Thanks for sharing these inputs, James. Based on that, here's what I propose:

#### Title
- Enable on the cart page
- Enable on the single product page
- Enable on the checkout page

#### Label
- Enable PayPal checkout button on the cart page
- Enable PayPal checkout button on the single product page
- Enable PayPal checkout button on the regular checkout page

#### Description (tool-tip)
I think we should remove tool-tips for all three of these because the title and label itself is getting repetitive and tool-tip will make us repeat these lines the third time.

@james-allan what do you think?

james-allan:
> what do you think?

Sounds good to me :) 

Just 2 little tweaks to the description text:

**_Enable PayPal Checkout buttons on the cart page_**

I've uppercased Checkout and pluralised buttons.

james-allan:
Finally getting around to review this again. The changes I requested in my last comment were addressed in 518225a. This LGTM.

Issue #759, opened 2020-05-30
Hiding PayPal Credit or Debit/Credit Cards as a Funding Method also hides Venmo button


fhaps:
### Describe the bug

If you chose to hide either PayPal Credit or Debit/Credit Cards as a Funding Method available to customers, the Venmo button also disappears.

When none of these two funding methods is hidden, the Venmo button is displayed correctly.

### To reproduce

1. Go to **WooCommerce > Settings > Payments > PayPal Checkout**
2. Under Hide Funding Methods, select PayPal Credit.
3. Go to the frontend of the site, add a product to the cart.
4. On the cart page, the smart payment buttons will have hidden the PayPal Credit button, but ALSO the Venmo button.

**Screenshots**

Without hiding any funding methods:

![](https://d.pr/i/nZJLLg+)
Image Link: https://d.pr/i/nZJLLg

After hiding PayPal Credit:

![](https://d.pr/i/H0fdVY+)
Image Link: https://d.pr/i/H0fdVY

### Expected behavior

For Venmo to appear as a funding method when it hasn't been hidden.

AashikP:
Setting priority so that this gets triaged. Also set to `Critical` for now since this disables a core feature offered by the extension.

We have a separate option to hide `Venmo`, so hiding PayPal Credit should not cause this behavior:

![https://d.pr/i/Q9sWF0+](https://d.pr/i/Q9sWF0+)

jorgeatorres:
Hey @fhaps @AashikP!

Thanks for the report. Are you using some kind of publicly available sandbox/test site to test this? We've been having trouble reproducing this problem, even when testing on a US mobile or using a US VPN. If this is on a test site you control, would it be possible for us to take a look?

Alternatively, could you repeat the test but keeping both PayPal Credit and Debit/Credit Cards enabled (that is, not in the "Hide Funding Method(s)" list) but instead hiding a different funding method (say, "ELV")?

That last test, in particular, can give us a lot of insight: while most of the time the logic used by PayPal to determine which buttons get displayed is outside of our control, I don't believe it is a coincidence that hiding a method (any method) always ends up hiding Venmo. If we're able to confirm this, then the issue can probably be explained by a layout problem, one that I'm trying to address with this PR: #774.

AashikP:
Hi @jorgeatorres  👋

I could not test the feature myself as I believe Venmo cannot be used from India (please do correct me if wrong)?

[...]

PR #774, opened 2020-06-17
Preserve layout style setting when using standalone buttons

jorgeatorres:
### Description

While investigating #759 I came across this issue: when the button layout is set to "Vertical" and at least one funding method needs to be hidden (including PayPal Credit), the buttons seem to "think" they're being rendered in a "horizontal" layout (as reflected by the CSS classes attached to the containers), even though they're actually being presented in a vertical one.

This is because we're rendering all eligible buttons using what PayPal calls "[standalone buttons](https://developer.paypal.com/docs/checkout/integration-features/standalone-buttons/)" and not passing the layout to the SDK except when taking care of the main PayPal button. We did this because not all style properties are supported by all buttons, but "layout" does seem to be supported in all scenarios.

This is probably a non-issue most of the time, but given that it is possible for funding methods to determine their eligibility by considering the layout they're in ([here's an example](https://github.com/paypal/paypal-checkout-components/blob/1b66936768c2430f41258537ae867904a612187a/src/funding/venmo/config.jsx#L48-L50)), it'd be better if we always explicitly indicate the layout.

### Steps to test:

1. Check out `master`.
1. Go to the extension settings (WC > Settings > Payment > PayPal Checkout):
   1. For simplicity, disable all context specific settings.
   1. Set "Button Layout" to "Vertical".
   1. Select at least one method inside the "Hide Funding Method(s)" field.
   1. Save changes.
1. Go the frontend, and add a product to the cart.
1. Visit the cart or checkout page.
1. Use your browser's devtools to inspect the "woo_pp_ec_button_cart" or the "woo_pp_ec_button_checkout" element (depending on which page you're looking at).
1. Confirm that all buttons (`<div role="button">`) except for the first one have a `paypal-buttons-layout-horizontal` CSS class. This is, despite the layout being "Vertical".
1. Check out this branch.
1. Reload the page and confirm that all buttons now have the correct CSS class: `paypal-buttons-layout-vertical`.

### DIFF
--- a/assets/js/wc-gateway-ppec-smart-payment-buttons.js
+++ b/assets/js/wc-gateway-ppec-smart-payment-buttons.js
@@ -222,7 +222,7 @@
 						onError:       button_args.onError,
 						onCancel:      button_args.onCancel,
 						fundingSource: fundingSource,
-						style:         ( paypal.FUNDING.PAYPAL === fundingSource ) ? button_args.style : {}
+						style:         ( paypal.FUNDING.PAYPAL === fundingSource ) ? button_args.style : { layout: button_args.style.layout }
 					};
 
 					var button = paypal.Buttons( buttonSettings );


mattallan:
Thanks @jorgeatorres and nice sleuthing!

I did a bunch of testing on this PR see https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/758#issuecomment-645718927 and code LGTM :+1:

Merring

PR #775, opened 2020-06-19
Make Mini-cart settings independent from cart

### LINKED ISSUE #745: Allow disabling the buttons on the mini-cart independently from the cart
### Describe the bug
We have a section with mini-cart specific settings, but whether the buttons are enabled or not on the mini-cart is controlled by the "Enable PayPal Checkout on the cart page" setting, which also controls the cart page.
We could look into adding a setting specifically for the mini-cart that is independent from the cart one.

Related to #744.

achyuthajoy:
### Description
At the moment, the PPEC on Mini-cart is controlled by the cart option. There are no specific options to disable Mini-cart PPEC and requires PPEC on cart to be enabled to show up.

This PR adds the following changes.

1. Re-purposes existing button settings as Global Settings, which takes effect if context-specific configuration is not defined.
2. Added context-specific configuration for Cart
3. Added option to disabled PPEC on Mini-cart
4. Re-purposes context-specific Mini-cart settings to show up only if PPEC on Mini-cart is enabled 

Tested the changes on a new site & existing site that already uses PPEC. Couldn't find any compatibility issues and noticed that the existing settings are carried over.

### Testing Changes
- [ ] Compare the PR against a new WooCommerce installation with the WordPress.org repository version to see if the default options are set correctly. PR will check enable mini-cart by default since there was no option to turn off PPEC on Mini-cart previously.
- [ ] On an existing site that uses PPEC, remove the plugin and use the development version to see if the previous settings are carried over. 
- [ ] Test context-specific settings for Single Product / Mini-cart / Cart / Checkout along with the option to disable PPEC on specific pages (especially Mini-cart since it's a new feature).

### Question

Since Mini-cart is enabled based on cart settings, on existing sites when the plugin is updated to include this PR, "Enable PayPal Checkout on Mini-cart" (default field value set to "yes") doesn't take effect if the settings are not saved.

Should we add an admin notice to save the settings once the plugin is updated? Or set the value of "Enable PayPal Checkout on Mini-cart" to "no" by default?
 
Not sure about documenting the new Global Settings & Mini-cart sections.

Closes #745.

### DIFF
--- a/assets/js/wc-gateway-ppec-settings.js
+++ b/assets/js/wc-gateway-ppec-settings.js
@@ -267,37 +267,57 @@
 			var display =
 				$( '#woocommerce_ppec_paypal_cart_checkout_enabled' ).is( ':checked' ) ||
 				( $( '#woocommerce_ppec_paypal_checkout_on_single_product_enabled' ).is( ':checked' ) && ! $( '#woocommerce_ppec_paypal_single_product_settings_toggle' ).is( ':checked' ) ) ||
-				( $( '#woocommerce_ppec_paypal_mark_enabled' ).is( ':checked' ) && ! $( '#woocommerce_ppec_paypal_mark_settings_toggle' ).is( ':checked' ) );
+				( $( '#woocommerce_ppec_paypal_mark_enabled' ).is( ':checked' ) && ! $( '#woocommerce_ppec_paypal_mark_settings_toggle' ).is( ':checked' ) ) ||
+				( $( '#woocommerce_ppec_paypal_cart_checkout_enabled').is( ':checked' ) && ! $( '#woocommerce_ppec_paypal_cart_settings_toggle' ).is( ':checked' ) ) ||
+				( $( '#woocommerce_ppec_paypal_mini_cart_checkout_enabled').is( ':checked' ) && ! $( '#woocommerce_ppec_paypal_mini_cart_settings_toggle' ).is( ':checked' ) );
 
 			$( '#woocommerce_ppec_paypal_button_layout, #woocommerce_ppec_paypal_button_size, #woocommerce_ppec_paypal_hide_funding_methods, #woocommerce_ppec_paypal_credit_enabled' ).closest( 'tr' ).toggle( display );
 			display && $( '#woocommerce_ppec_paypal_button_layout' ).change();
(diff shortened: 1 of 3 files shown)


jorgeatorres (changes requested):
Hi @achyuthajoy!

Thanks for submitting this 🙇! I've left some feedback, which mostly deals with minor things. You did an awesome job capturing the requirement and proposing a solution, even though it involved navigating very different parts (settings, JavaScript, PHP) 💯.

Somethings I noticed and that match some of your questions/observations:

- I don't think we should add context-specific configuration for the cart at this point. I know the UI isn't ideal and what you did here works perfectly fine, but that the cart and "default" settings are shared is something that has been that way for a long time… and so we could be adding even more confusion. Also, it's not clear what the defaults mean if you enable context specific settings for all contexts. In contrast, doing context specific configuration for other contexts (while more work) does mean the only context remaining (cart) uses the default, which is the current behavior.
   - We could look into clarifying all of this in a different PR, while we concentrate on separating the mini-cart toggle in this one. We can also discuss how to migrate to the new settings, if necessary.
- As you yourself have pointed out, when upgrading things don't work correctly. You see the mini-cart enabled but it really isn't. Instead of adding an admin notice we could look into automatically (and silently) populating the settings accordingly. The value for this new setting (`mini_cart_checkout_enabled`) should be the same as for `cart_checkout_enabled`.
   We already do something like this [here](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/3d3a7965c7b97f1f7a043e9d2f095a6f364d1dae/includes/class-wc-gateway-ppec-settings.php#L145). You can use that as inspiration, but basically, we would set `$_settings['mini_cart_checkout_enabled']` to the value of the cart setting if it is not already present in the array. I think this is the right way, but alternatively, you could just leave the "yes" as default and reverse some of the [conditional checks](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/775/files#diff-c5271c5e9f4b2ddfc403c66a5d3c5375R418) so that they do not show the PayPal button in the mini cart when explicitly set to "no".

Once again, good job here 👍. Feel free to respond to the feedback in case you have questions or want to continue the discussion.

Issue #416, opened 2018-05-23
Header image not showing at full width


hannahswain:
Somewhat related to this issue: https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/371

When you set the PayPal header image to an image of 750x90:

![PayPal Settings](http://cld.wthms.co/kBwXCt+)
Full size: http://cld.wthms.co/kBwXCt

It does show up on the PayPal screen, but cropped as if it were the 190x60 image instead.

![PayPal checkout](http://cld.wthms.co/xHFQZB+)
Full size: http://cld.wthms.co/xHFQZB

coderkevin:
I couldn't get the image to show up at the top of the paypal checkout at all, maybe they don't do that anymore or maybe it's a page style issue?

However, on the admin page, it looks like nothing has changed here.

![WooCommerce_settings_‹_paypal-express_—_WordPress](https://user-images.githubusercontent.com/945228/68423071-04ae1c00-0167-11ea-9b1a-44b6f56a33d2.png)

v18:
I'm able to repro this:

<img width="656" alt="Screen Shot 2019-11-12 at 12 19 49 AM" src="https://user-images.githubusercontent.com/11487924/68644197-3a922e00-04e2-11ea-9082-cf91fcebb981.png">

PR #779, opened 2020-06-24
Display correct image size in the PayPal Checkout window

codestor4:
## Summary

This fixes https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/416. 

- This defines image sizes for `ppec_logo_image_size` as 190x60 and `ppec_header_image_size` as 750x90 using [add_image_size](https://developer.wordpress.org/reference/functions/add_image_size/) function.
- Then renders the image based on this condition: `'logo_image_url' === $key ? 'ppec_logo_image_size' : 'ppec_header_image_size' `.

## Steps to test:

1. Switch to the master/main branch.
1. Use the browser in full-screen (otherwise PayPal window will be too short to display the logo image).
1. Checkout using PPEC, you will see a cropped image.
1. Switch to the branch `issue_416`, try checking out again, you will see the image in proper width. Checkout the screenshots below for a before and after.

## Documentation

- This PR does not need documentation.

## Questions

- Have I added the `add_image_size` function in the right place? Or is there a better place for it?
- Not directly related to the issue: When the browser isn't in full-screen mode, checking out will display a small PayPal window, hiding the image logo. We only see the logo when we resize the window or try checking out in full-screen mode. Is this intentional? If it isn't, any advice how could we increase the width and height of the PP popup window?

## Before the fix

Screenshot 2020-06-24 at 10 32 23 PM

## After the fix

Screenshot 2020-06-24 at 10 32 55 PM

### DIFF
--- a/includes/abstracts/abstract-wc-gateway-ppec.php
+++ b/includes/abstracts/abstract-wc-gateway-ppec.php
@@ -559,6 +559,11 @@ abstract class WC_Gateway_PPEC extends WC_Payment_Gateway {
 		// For backwards compatibility (customers that already have set a url)
 		$value_is_url = filter_var( $value, FILTER_VALIDATE_URL ) !== false;
 
+		if ( function_exists( 'add_image_size' ) ) {
+			add_image_size( 'ppec_logo_image_size', 190, 60 );
+			add_image_size( 'ppec_header_image_size', 750, 90 );
+		}
+
 		if ( empty( $value ) || $value_is_url ) {
 			$maybe_hide_remove_style = 'display: none;';
 		} else {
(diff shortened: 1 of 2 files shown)


includes/abstracts/abstract-wc-gateway-ppec.php
```diff
@@ -559,6 +559,11 @@ public function generate_image_html( $key, $data ) {
 		// For backwards compatibility (customers that already have set a url)
 		$value_is_url = filter_var( $value, FILTER_VALIDATE_URL ) !== false;
 
+		if ( function_exists( 'add_image_size' ) ) {
```
jorgeatorres, line 562: To answer one of your questions, while this seems to work here, I think the constructor would be a better place for these `add_image_size()` calls.

codestor4:
>  I left some very minor feedback. Other than that, this works perfectly fine.

Thank you :) 

> PayPal documentation says that the header image field (HDRIMG) has been deprecated (with no alternative). We could probably drop it entirely but that's something for a different time/issue.

I didn't think of that. Thanks for sharing this detail. I agree it could be done in separate issue.

> With your fixes, if the window gets to a size in which the logo is displayed, it'll look fine, which I believe addresses the original issue.

Yes, it does. I'll make another commit to address your inline feedback. Then we can merge and close this.

includes/abstracts/abstract-wc-gateway-ppec.php
codestor4: Done. Tested. Works. ✅

PR #781, opened 2020-06-29
Add filters 

ahegyes:
### Description

Added filters for retrieving the line items from cart and from orders. 

We stumbled upon a situation where we needed to alter the name of the line items sent to PayPal to take into account custom product fields added with another plugin. Just using the filter inside of "product->get_name()" is not enough since we needed to include the information stored in the custom product fields inside the actual line items. 

This was possible using WC's built-in PayPal integration (see filters "woocommerce_paypal_get_order_item_name" and "woocommerce_paypal_line_item" inside WooCommerce). This PR ensures parity in this regard with the built-in PayPal gateway.

### Steps to test:

This doesn't make any changes to the way the plugin works. It just enables users to filter the line items sent to PayPal to include their own custom logic for generating them.

### Documentation

Customers won't notice any changes.

### DIFF
--- a/includes/class-wc-gateway-ppec-client.php
+++ b/includes/class-wc-gateway-ppec-client.php
@@ -515,7 +515,7 @@ class WC_Gateway_PPEC_Client {
 				'sku'         => $values['data']->get_sku(),
 			);
 
-			$items[] = apply_filters( 'woocommerce_paypal_express_get_line_item_from_cart', $item, $values, $cart_item_key );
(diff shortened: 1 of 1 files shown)


ahegyes:
Hello! Sorry to intervene, but I was wondering if there are any plans to merge this request or if this might've been rejected? Thank you!

jorgeatorres:
Hi @ahegyes,

Thanks for taking the time to contribute! We apologize for the delay in responding.
The filters look like a nice addition. My only concern was that they make the whole line item susceptible of being altered. Based on your PR and with the goal of limiting modifications to line item name/description, I've submitted PR #869. Hopefully these new filters will also work for your use case 😊.

PR #783, opened 2020-06-30
Simplify Travis configuration

jorgeatorres:
Now that our codebase is in better shape re standards compliance, we no longer need to have 2 jobs running PHP Code Sniffer on every PR. This PR simplifies our Travis script to reflect the new situation.


mattallan:
Thanks @jorgeatorres

PR #786, opened 2020-07-06
WIP: Setup E2E testing using @woocommerce/e2e-environment

mattallan:
This PR will act as a WIP for setting up the e2e tests and writing our first couple of e2e tests.

**Changes proposed in this PR**
 - Sets up e2e test environment using the `@woocommerce/e2e-environment` package
   - I followed the instructions found in the [WC e2e env README](https://github.com/woocommerce/woocommerce/blob/master/tests/e2e/env/README.md) to set this up.
 - Adds a readme with testing instructions and running the tests locally
 - Adds our first example E2E test which just loads the checkout page 🎉

Other things to note:
 - While WooCommerce is still working on a `@woocommerce/e2e-util` package, I have copied the util files from core into our own `utils` folder, see f7a3473.
 - The default docker init script was just installing and activating WooCommerce and the 2019 theme, so to help with testing, I have modified this script to be similar some of what is done in WC Payments. See 8cb68a4.

### How to run the e2e tests?

Check out the [README added in this PR](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/blob/0081f8faa552405c7e19219ca3f53b36745d3389/tests/e2e/README.md) which has more detailed instructions.

1. Install docker and make sure it's running
1. `npm install`
2. `npm run docker:up`
   1. This command is booting up the test environment, which includes installing and setting up WooCommerce and importing some dummy/test products. This happens in the background so it can take a while to finish even though the `docker:up` command has completed.
   2. To check that the site is ready for testing and set up properly, go to http://localhost:8084/shop/ and confirm the shop loads with the storefront theme and there are products in your store. It should look like [this](https://d.pr/i/1bevdM).
3. Run the example test: `npm run test:e2e`

4. To watch the tests running in a chrome window you can use `npm run test:e2e-dev`
5. `npm run docker:down` when finished testing

**TODOs**
At the moment this PR is for running the e2e tests locally

 - [ ] Travis CI integration/setup


mattallan:
At the start of every test I've currently got it importing calling [`await importSampleProducts()`](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/786/files#diff-398e5c81d9b4ac0b2e8ebaf5cda2c778R246) to make sure our shop page is populated but I'd like to find a better way of doing this.

I originally had the importing of sample products using wp CLI which I changed in 23589ee because Travis was running the e2e tests before the `initialize.sh` had time to finish importing the products so the tests would fail.

I think we'll need to look at how other extensions handle setting up the test environment using WPI CLI

PR #794, opened 2020-08-19
Use Lambda team's standard github README and templates

chickenn00dle:
### Description

PPEC currently has no README. This PR adds one following the standard from our other repos.

This also updates the PR template to include a changelog entry.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -24,4 +24,7 @@
 - [ ] This PR needs documentation (has the "Documentation" label).
 <!-- For an extra 💯 include further details about which change requires documentation -->
 
+### Changelog entry
+> Enter a summary of all changes on this Pull Request. This will appear in the changelog if accepted.
+
 Closes # .
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ A "_deployment_" in our sense means:
 
 ## Branches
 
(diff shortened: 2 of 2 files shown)


LCmry:
Looks good @chickenn00dle , one small change is that the `main` branch is now `trunk` 😅

chickenn00dle:
Ah, good catch @LCmry!

I updated the README in https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/794/commits/52d3a8fafa8241c2acd47408ebf87606e7d4fce2

LCmry:
One more Q @chickenn00dle , where is the Changelog entry being drawn from. AFAIK it's not in our other repos, at least unless we inherited it. I know this is a practice other teams do, I wonder what the pros and cons are and if it will be helpful at release time?

chickenn00dle:
@LCmry This was discussed when we standardized the Square templates:
https://github.com/woocommerce/woocommerce-square/pull/470

You can find specific details in this comment:
https://github.com/woocommerce/woocommerce-square/pull/470#discussion_r455470463

I believe there may have also been a Slack discussion around this, but am not finding it atm.

Let me know if this doesn't answer your question!

LCmry:
Gotcha, thanks for the trail @chickenn00dle! LGTM 👍

PR #797, opened 2020-08-26
Replace instances of main with trunk

chickenn00dle:
This replaces instances of main with trunk for the repo per pb0GrA-R4-p2


chickenn00dle:
Merging as this is basically a repeat of #792

PR #798, opened 2020-08-26
Replaces instances of main with trunk from main

chickenn00dle:
This replaces references to main with trunk per pb0GrA-R4-p2

These changes were accidentally applied to `main` branch before I set `trunk` as default. So this gets these changes added to `trunk` as well 😅

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ php:
 
 branches:
   only:
-    - main
+    - trunk
 
 install:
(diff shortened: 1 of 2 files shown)


chickenn00dle:
Leaving this one up for review simply because of the mixup with merging to main initially.

Nutza97 (approved):

jorgeatorres (approved):

PR #837, opened 2020-11-19
Use is_callable() instead of method_exists() in check

### LINKED ISSUE #829: Confirm plugin is compatible with upcoming PHP 8
PHP 8.0 is planned for November 26th. We need to test and make sure the plugin works with this version.

See https://github.com/php/php-src/blob/107962208a19d8b6dc1a190cb25fc37614411e71/UPGRADING.

jorgeatorres:
**Issue**: #829

### Description

After doing some [code analysis and smoke tests](https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/issues/829) I didn't notice any major problems with PHP 8 and PPEC.

The only change I'm pushing modifies a check so that it uses `is_callable()` instead of `method_exists()`. This is probably not necessary but better to err on the side of caution.

Closes #829.

### DIFF
--- a/includes/class-wc-gateway-ppec-checkout-handler.php
+++ b/includes/class-wc-gateway-ppec-checkout-handler.php
@@ -135,7 +135,7 @@ class WC_Gateway_PPEC_Checkout_Handler {
 			return $fields;
 		}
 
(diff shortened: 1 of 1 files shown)


jorgeatorres:
Self-merging as this is an innocuous change.

PR #856, opened 2021-02-23
Avoid fatal error when line_item is not an instance of WC_Product

mircobabini:
### Description
Sometimes plugins and snippets out there uses this code to add custom lines to the cart (not instance of WC_Product):
``` WC()->cart->add_to_cart( array );```

This is causing a fatal error on `$product->get_sku()`.

### Steps to test:
1. For instance, install LMS Masterstudy (free plugin from WordPress.org)
1. In options activate WC Checkout; then create a course and add it to the cart
1. Just complete the checkout and boom! Fatal error instead of the thank you page.

### Documentation

Closes nothing; issue not reported, preferred to send this pull request.


jorgeatorres:
Closed in favor of #872.

PR #865, opened 2021-05-19
Adds an option for displaying WooCommerce Inbox notifications

mattallan:
Internal link: paJDYF-1uJ-p2
Parent GH link: 10117-gh-Automattic/woocommerce.com

---

### Description

This PR adds the WC Admin inbox notification option to PPXO as per: pb0GrA-1cH-p2#comment-2063

### Steps to test:
The option should be added on plugin activation and upgrade.

**Plugin Activation**
1. Checkout branch `wc_inbox_option`
2. Deactivate PayPal Checkout plugin
3. Delete `woocommerce_inbox_variant_assignment` from wp_options if it already exists
4. Activate PayPal Checkout
5. Check `woocommerce_inbox_variant_assignment` is a random value from 1-12

**Plugin Upgrade**

1. Delete `woocommerce_inbox_variant_assignment` from wp_options if it already exists
2. Go into wp_options table and edit `wc_ppec_version` to be 1 version behind (i.e. 2.1.0)
3. Reload an admin page to trigger an update
4. Check `woocommerce_inbox_variant_assignment` is a random value from 1-12

### DIFF
--- a/woocommerce-gateway-paypal-express-checkout.php
+++ b/woocommerce-gateway-paypal-express-checkout.php
@@ -47,3 +47,19 @@ function wc_gateway_ppec() {
 }
 
 wc_gateway_ppec()->maybe_run();
+
+/**
+ * Adds the WooCommerce Inbox option on plugin activation
+ *
+ * @since 2.1.2
+ */
+if ( ! function_exists( 'add_woocommerce_inbox_variant' ) ) {
+	function add_woocommerce_inbox_variant() {
+		$option = 'woocommerce_inbox_variant_assignment';
+
+		if ( false === get_option( $option, false ) ) {
+			update_option( $option, wp_rand( 1, 12 ) );
+		}
+	}
+}
+register_activation_hook( __FILE__, 'add_woocommerce_inbox_variant' );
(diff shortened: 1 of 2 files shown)


woocommerce-gateway-paypal-express-checkout.php
```diff
@@ -47,3 +47,19 @@ function wc_gateway_ppec() {
...
 wc_gateway_ppec()->maybe_run();
+
+/**
+ * Adds the WooCommerce Inbox option on plugin activation
+ *
+ * @since 2.1.2
+ */
+if ( ! function_exists( 'add_woocommerce_inbox_variant' ) ) {
+	function add_woocommerce_inbox_variant() {
+		$option_name = 'woocommerce_inbox_variant_assignment';
```
LCmry: Similar to what was mentioned here: https://github.com/woocommerce/woocommerce-paypal-payments/pull/160#pullrequestreview-655255648
This should be `$option`.

LCmry:
Thanks @mattallan. Other than the variable name, this looks good. I tested both scenarios and it worked as expected.

mattallan:
Thanks for the review @LCmry !
I've just forced pushed that variable name preference 😃 This should be good for another review, thanks!

PR #866, opened 2021-05-20
Upgrade to PayPal Payments Notice on Plugins Page

achyuthajoy:
### Description

As part of various efforts to drive adoption of PayPal Payments, this PR aims to add a notice on the WordPress plugins page suggesting users to upgrade from PayPal Checkout to PayPal Payments.

**Screenshot of the notice**

Screenshot

**Screencast** - https://d.pr/v/Y24RUS

### Steps to test:
1. Checkout the branch `paypal-payments-upgrade-notice` & activate the plugin if not active.
2. On the installed plugins page, a new section is added just below PayPal Checkout suggesting upgrade to PayPal Payments.

- [ ] Confirm that `Learn more` button opens up product documentation on WooCommerce.com
- [ ] Confirm that `Upgrade to PayPal Payments now` button installs the plugins and redirects to activation page.
- [ ] Confirm that when `Upgrade to PayPal Payments now` button is clicked, the button text changes to `Installing...` with a spinning icon.
- [ ] Confirm that the button text is `Upgrade to PayPal Payments now` if plugin is not installed, `Activate PayPal Payments now` if plugin is installed but not active.
- [ ] Confirm that when the plugin is deleted, the notice button immediately changes to `Upgrade to PayPal Payments now` instead of `Activate PayPal Payments now`.

The SCSS code used can be found here - https://gist.github.com/achyuthajoy/06a2012495d2d2816678fe15e9860c05

Adds notice on plugins page to upgrade to PayPal Payments.

### DIFF
--- a/assets/js/admin/ppec-upgrade-notice.js
+++ b/assets/js/admin/ppec-upgrade-notice.js
@@ -8,23 +8,21 @@
 	}
 
 	// Handle delete event for PayPal Payments.
-	$( document ).on( 'wp-plugin-delete-success', function( response ) {
-		setTimeout( function(){
-			if ( is_paypal_payments_installed && $( 'tr#woocommerce-paypal-payments-deleted' ).length ) {
-				is_paypal_payments_installed = false;
+	$( document ).on( 'wp-plugin-delete-success', function( event, response ) {
+		if ( is_paypal_payments_installed && 'woocommerce-paypal-payments' === response.slug ) {
+			is_paypal_payments_installed = false;
 
-				// Change PPEC notice activation button id, text & link.
-				const ppec_install_id   = $( '#ppec-activate-paypal-payments' ).data( 'install-id' );
-				const ppec_install_text = $( '#ppec-activate-paypal-payments' ).data( 'install-text' );
-				const ppec_install_link = $( '#ppec-activate-paypal-payments' ).data( 'install-link' );
+			// Change PPEC notice activation button id, text & link.
+			const ppec_install_id   = $( '#ppec-activate-paypal-payments' ).data( 'install-id' );
+			const ppec_install_text = $( '#ppec-activate-paypal-payments' ).data( 'install-text' );
+			const ppec_install_link = $( '#ppec-activate-paypal-payments' ).data( 'install-link' );
 
-				$( '#ppec-activate-paypal-payments' ).text( ppec_install_text );
-				$( '#ppec-activate-paypal-payments' ).attr({
-					href: ppec_install_link,
-					id: ppec_install_id
-				});
-			}
-		}, 500 );
+			$( '#ppec-activate-paypal-payments' ).text( ppec_install_text );
+			$( '#ppec-activate-paypal-payments' ).attr({
+				href: ppec_install_link,
+				id: ppec_install_id
+			});
+		}
 	} );
 
 	// Change button text when install link is clicked.
--- a/assets/js/admin/ppec-upgrade-notice.js
+++ b/assets/js/admin/ppec-upgrade-notice.js
@@ -12,6 +12,10 @@
 		if ( targetElement.hasClass( 'active' ) ) {
 			is_paypal_payments_active = true;
 		}
+
+		// Dynamically update plugin activation link to handle plugin folder renames.
+		let activation_url = $( targetElement ).find( 'span.activate a' ).attr( 'href' );
+		$( 'a#ppec-activate-paypal-payments' ).attr( 'href', activation_url );
 	}
 
 	// Hide notice/buttons conditionally.
--- a/assets/js/admin/ppec-upgrade-notice.js
+++ b/assets/js/admin/ppec-upgrade-notice.js
@@ -2,16 +2,26 @@
 	'use strict';
 
 	// Check whether PayPal Payments is installed.
-	let is_paypal_payments_installed = false;
-	if ( $( 'tr[data-slug="woocommerce-paypal-payments"]' ).length ) {
+	let is_paypal_payments_installed = false,
+		is_paypal_payments_active = false;
+
+	const targetElement = $( 'tr[data-slug="woocommerce-paypal-payments"]' );
+	if ( targetElement.length ) {
 		is_paypal_payments_installed = true;
+
+		if ( targetElement.hasClass( 'active' ) ) {
+			is_paypal_payments_active = true;
+		}
+	}
+
+	// Hide notice is PayPal Payments is installed and active.
+	if ( is_paypal_payments_installed && is_paypal_payments_active ) {
+		$( 'tr#ppec-migrate-notice' ).hide();
 	}
 
 	// Handle delete event for PayPal Payments.
 	$( document ).on( 'wp-plugin-delete-success', function( event, response ) {
 		if ( is_paypal_payments_installed && 'woocommerce-paypal-payments' === response.slug ) {
-			is_paypal_payments_installed = false;
-
 			// Change PPEC notice activation button id, text & link.
 			const ppec_install_id   = $( '#ppec-activate-paypal-payments' ).data( 'install-id' );
 			const ppec_install_text = $( '#ppec-activate-paypal-payments' ).data( 'install-text' );
--- a/assets/js/admin/ppec-upgrade-notice.js
+++ b/assets/js/admin/ppec-upgrade-notice.js
@@ -14,24 +14,20 @@
 		}
 	}
 
-	// Hide notice is PayPal Payments is installed and active.
+	// Hide notice/buttons conditionally.
 	if ( is_paypal_payments_installed && is_paypal_payments_active ) {
 		$( 'tr#ppec-migrate-notice' ).hide();
+	} else if ( is_paypal_payments_installed ) {
+		$( 'a#ppec-install-paypal-payments' ).hide();
+	} else {
+		$( 'a#ppec-activate-paypal-payments' ).hide();
 	}
 
 	// Handle delete event for PayPal Payments.
 	$( document ).on( 'wp-plugin-delete-success', function( event, response ) {
 		if ( is_paypal_payments_installed && 'woocommerce-paypal-payments' === response.slug ) {
-			// Change PPEC notice activation button id, text & link.
-			const ppec_install_id   = $( '#ppec-activate-paypal-payments' ).data( 'install-id' );
-			const ppec_install_text = $( '#ppec-activate-paypal-payments' ).data( 'install-text' );
-			const ppec_install_link = $( '#ppec-activate-paypal-payments' ).data( 'install-link' );
-
-			$( '#ppec-activate-paypal-payments' ).text( ppec_install_text );
-			$( '#ppec-activate-paypal-payments' ).attr({
-				href: ppec_install_link,
-				id: ppec_install_id
-			});
+			$( 'a#ppec-activate-paypal-payments' ).hide();
+			$( 'a#ppec-install-paypal-payments' ).show();
 		}
 	} );
 
--- a/assets/js/admin/ppec-upgrade-notice.js
+++ b/assets/js/admin/ppec-upgrade-notice.js
@@ -27,6 +27,9 @@
 		$( 'a#ppec-activate-paypal-payments' ).hide();
 	}
 
+	// Display buttons area
+	$( '#ppec-migrate-notice .ppec-notice-buttons' ).removeClass( 'hidden' );
+
 	// Handle delete event for PayPal Payments.
 	$( document ).on( 'wp-plugin-delete-success', function( event, response ) {
 		if ( is_paypal_payments_installed && 'woocommerce-paypal-payments' === response.slug ) {
--- a/assets/js/admin/ppec-upgrade-notice.js
+++ b/assets/js/admin/ppec-upgrade-notice.js
@@ -1,0 +1,39 @@
+;(function ( $, window, document ) {
+	'use strict';
+
+	// Check whether PayPal Payments is installed.
+	let is_paypal_payments_installed = false;
+	if ( $( 'tr[data-slug="woocommerce-paypal-payments"]' ).length ) {
+		is_paypal_payments_installed = true;
+	}
+
+	// Handle delete event for PayPal Payments.
+	$( document ).on( 'wp-plugin-delete-success', function( response ) {
+		setTimeout( function(){
+			if ( is_paypal_payments_installed && $( 'tr#woocommerce-paypal-payments-deleted' ).length ) {
+				is_paypal_payments_installed = false;
+
+				// Change PPEC notice activation button id, text & link.
+				const ppec_install_id   = $( '#ppec-activate-paypal-payments' ).data( 'install-id' );
+				const ppec_install_text = $( '#ppec-activate-paypal-payments' ).data( 'install-text' );
+				const ppec_install_link = $( '#ppec-activate-paypal-payments' ).data( 'install-link' );
+
+				$( '#ppec-activate-paypal-payments' ).text( ppec_install_text );
+				$( '#ppec-activate-paypal-payments' ).attr({
+					href: ppec_install_link,
+					id: ppec_install_id
+				});
+			}
+		}, 500 );
+	} );
+
+	// Change button text when install link is clicked.
+	$( document ).on( 'click', '#ppec-install-paypal-payments', function( e ) {
+		e.preventDefault();
+		$( this ).addClass( 'updating-message' ).text( 'Installing...' );
+		const install_link = $( this ).attr('href');
+		setTimeout( function(){
+			window.location = install_link;
+		}, 100 );
+	});
+})( jQuery, window, document );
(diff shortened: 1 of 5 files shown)


achyuthajoy:
@jorgeatorres Not sure why Travis is failing. The error doesn't come up when I'm running PHPCS locally with php@7.4. Travis is running php@7.0

achyuthajoy:
The same error showed up locally when I switched to php@7.0. But there is no error with the code. Weird.

achyuthajoy:
@jorgeatorres The error with php@7.0 is fixed now. All good.

includes/class-wc-gateway-ppec-plugin.php
```diff
@@ -492,6 +494,28 @@ public static function needs_shipping() {
...
+	 * Displays notice to upgrade to PayPal Payments.
+	 *
+	 * @param string $plugin_file Path to the plugin file relative to the plugins directory.
+	 * @param array $plugin_data An array of plugin data.
+	 * @param string $status Status filter currently applied to the plugin list.
+	 */
+	public function ppec_upgrade_notice( $plugin_file, $plugin_data, $status ) {
+		// Return if not on installed plugins page.
+		global $pagenow;
+		if ( 'plugins.php' !== $pagenow ) {
```
jorgeatorres: Is this check really needed? Could the `after_plugin_row_(plugin)` action be fired on other pages?

includes/class-wc-gateway-ppec-plugin.php
```diff
@@ -492,6 +494,28 @@ public static function needs_shipping() {
...
+			return;
+		}
+
+		// Load styles & scripts required for the notice.
+		wp_enqueue_style( 'ppec-upgrade-notice', plugin_dir_url( __DIR__ ) . '/assets/css/admin/ppec-upgrade-notice.css', array(), WC_GATEWAY_PPEC_VERSION );
+		wp_enqueue_script( 'ppec-upgrade-notice-js', plugin_dir_url( __DIR__ ) . '/assets/js/admin/ppec-upgrade-notice.js', array(), WC_GATEWAY_PPEC_VERSION, false );
+
+		// Load notice template.
+		include_once $this->plugin_path . 'templates/paypal-payments-upgrade-notice.php';
```
jorgeatorres, line 510: Nice overall approach with this function. I like it 💯.

assets/js/admin/ppec-upgrade-notice.js
```diff
@@ -0,0 +1,39 @@
...
+	// Check whether PayPal Payments is installed.
+	let is_paypal_payments_installed = false;
+	if ( $( 'tr[data-slug="woocommerce-paypal-payments"]' ).length ) {
+		is_paypal_payments_installed = true;
+	}
+
+	// Handle delete event for PayPal Payments.
+	$( document ).on( 'wp-plugin-delete-success', function( response ) {
+		setTimeout( function(){
+			if ( is_paypal_payments_installed && $( 'tr#woocommerce-paypal-payments-deleted' ).length ) {
```
jorgeatorres: I might be wrong, but apparently `response` includes information about the deleted plugin(s). Would it be possible to use that instead of this check here? If it's too much trouble, we could leave it as is and try to find the "deleted" row to detect the deletion.

woocommerce-gateway-paypal-express-checkout.php
```diff
@@ -29,6 +29,9 @@
 
 define( 'WC_GATEWAY_PPEC_VERSION', '2.1.1' );
 
+// Define plugin base
+define( 'WC_GATEWAY_PPEC_PLUGIN_BASE', plugin_basename( __FILE__ ) );
```
jorgeatorres: Nothing against this in particular, but we're introducing a new global that is only used in one place.
Couldn't we just "assume" the path is `woocommerce-gateway-paypal-express-checkout/woocommerce-gateway-paypal-express-checkout.php` above?
This only seems useful if someone has renamed the plugin, but that seems unlikely and not necessarily something we should support. Thoughts?

Is there a particular reason you want to define a new global with the file name? Couldn't we

templates/paypal-payments-upgrade-notice.php
```diff
@@ -0,0 +1,110 @@
...
+	$button_data['href'] = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=ppcp-gateway' );
+	$button_data['text'] = 'Configure PayPal Payments now';
+
+} elseif ( $is_installed_paypal_payments ) {
+
+	// PayPal Payments installed, but inactive.
+	$button_data['id']         = 'ppec-activate-paypal-payments';
+	$button_data['href']       = $paypal_payments_activate_link;
+	$button_data['text']       = 'Activate PayPal Payments now';
+	$button_data['attributes'] = array(
```
jorgeatorres: I was thinking about this part specifically, and while every button only cares about its own id/href/text, this button in particular needs some context (which, in turn, is the id/href/text for another one). Because of that, it might be less complicated if we just print *both* buttons ("Activate" and "Upgrade") in the HTML and use the JS to hide/show the buttons accordingly.

What do you think? This could also slightly reduce some of the logic in the JS and the template itself, and also some of the "duplication" of the `$button_data` here for the `$is_installed_paypal_payments = true` case.

templates/paypal-payments-upgrade-notice.php
```diff
@@ -0,0 +1,110 @@
...
+				<p>Upgrade now and get access to these great features:</p>
+
+				<ul>
+					<li>Give your customers their preferred ways to pay with one checkout solution. Accept <strong>PayPal</strong>, <strong>PayPal Credit</strong>, <strong>Pay Later</strong> options (available in the US, UK, France, and Germany), <strong>credit & debit cards</strong>, and country-specific, <strong>local payment methods</strong> on any device.</li>
+					<li>Offer subscriptions and accept recurring payments as PayPal is compatible with <a target="_blank" href="https://woocommerce.com/products/woocommerce-subscriptions/"><strong>WooCommerce Subscriptions</strong></a>.</li>
+				</ul>
+			</div>
+			<div class='ppec-notice-buttons ppec-notice-section'>
+				<?php
+				$extra_attributes = '';
```
jorgeatorres: If we do the above, and print both buttons, we wouldn't need to add "extra attributes", making this part no longer necessary, including the anonymous function.

jorgeatorres (commented):
Hi @achyuthajoy!

Awesome work on all of this 💯. I love the approach you took in general. It's very clean and clever 😸 .

I didn't find any serious blockers, but I did leave some feedback related to certain implementation details. I'd like to read your thoughts on that.

- - -

On a not-code-related note, I'd like to discuss how the notice behaves when you've already installed and activated PayPal Payments:

Screen Shot 2021-06-09 at 21 38 03

It feels odd to me that once a merchant has both installed and activated PayPal Payments, we're still telling them about the upgrade to PayPal Payments in a very hard to miss notice (despite the fact that they've already read and acted on our messaging).

Couldn't we reduce the messaging somehow when PayPal Payments is already installed and active? If not, that's ok, but maybe we could hide the "Learn More" button or make the "Configure PayPal Payments now" a secondary button (like "Learn More" itself) and relabel as "Configure PayPal Payments" or something similar. The combination of the button being a primary button *and* the "Configure PayPal Payments **now**" label makes it look as if I (as a merchant) have missed something and still need to setup PayPal Payments, which is probably not the case.

includes/class-wc-gateway-ppec-plugin.php
achyuthajoy: Yeah you're right. It doesn't fire on other admin pages. I'll remove the same

includes/class-wc-gateway-ppec-plugin.php
achyuthajoy: 👍 I wanted to keep everything tied to less number of files / functions just in case we want to remove it later.

assets/js/admin/ppec-upgrade-notice.js
achyuthajoy: The `response` is supposed to return the ID of the plugin deleted. However, the data parameter returns `undefined`. Hence I used this logic.

Here is a screenshot of the response object.

Response

woocommerce-gateway-paypal-express-checkout.php
achyuthajoy: I added it for 2 reasons.

1. It's one of the recommended methods to get the plugin path.
2. Should work even of the plugin folder is renamed, which can happen.

> This only seems useful if someone has renamed the plugin, but that seems unlikely and not necessarily something we should support

Given that we have a way to support even if the plugin folder is renamed, I think we should keep it.

templates/paypal-payments-upgrade-notice.php
achyuthajoy: We can go with your suggestion to add all buttons and hide specific ones via JS. However, we would need to add HTML for for 4 buttons. But yes, it would reduce the logic added on the template.

1. Install
2. Activate
3. Configure
4. Learn More

If you think this is better, I'm happy to make the changes. I really don't have a preference in this case.

templates/paypal-payments-upgrade-notice.php
achyuthajoy: Sure. Based on your decision on the above thread, I'll make the changes as required.

achyuthajoy:
> It feels odd to me that once a merchant has both installed and activated PayPal Payments, we're still telling them about the upgrade to PayPal Payments in a very hard to miss notice (despite the fact that they've already read and acted on our messaging).
> 
> Couldn't we reduce the messaging somehow when PayPal Payments is already installed and active? If not, that's ok, but maybe we could hide the "Learn More" button or make the "Configure PayPal Payments now" a secondary button (like "Learn More" itself) and relabel as "Configure PayPal Payments" or something similar. The combination of the button being a primary button _and_ the "Configure PayPal Payments **now**" label makes it look as if I (as a merchant) have missed something and still need to setup PayPal Payments, which is probably not the case.

@LevinMedia Can you look into this and suggest options?

woocommerce-gateway-paypal-express-checkout.php
jorgeatorres: Fair enough. My point was that we don't do the same to detect PayPal Payments or something so maybe it shouldn't be _that_ necessary. Having said that, there's a `$file` public var in `WC_Gateway_PPEC_Plugin` that we could use instead of adding a new global. Could you make the change to use that?

templates/paypal-payments-upgrade-notice.php
jorgeatorres: "Learn More" is already in the HTML, so that wouldn't be too different. As for the others, we don't need to actually _print_ all of those 3 all the time (Install and Activate, for example).

If that's ok with you, I do like the idea of printing the HTML 👍

assets/js/admin/ppec-upgrade-notice.js
jorgeatorres: Yes, the event doesn't include the data, but the second argument should. The callback could be something like this:

```js
	$( document ).on( 'wp-plugin-delete-success', function( ev, response ) {
		// `response` has the details.
	} );
```

If you can, please give it a try. If that still doesn't work for you or using the `response` objects proves to be too complicated, we can keep the current approach. Let me know 😸 .

achyuthajoy: Sure. I'll give it a try @jorgeatorres. 

Based on this, I only see one param - https://github.com/WordPress/wordpress-develop/blob/2382765afa36e10bf3c74420024ad4e85763a47c/src/js/_enqueues/wp/updates.js#L1057

jorgeatorres: Yes, that's how `.trigger()` works. Because the parameter is not an event object, it is passed as an extra parameter. The callback, though, always receives the event and _then_ the rest of the extra parameters.

achyuthajoy: Gotcha! Will try that :D

woocommerce-gateway-paypal-express-checkout.php
achyuthajoy: I totally missed that. `$file` works. I've added a commit for the same.

LevinMedia:
@achyuthajoy I agree, once PayPal Payments is installed, I don't think we should show any additional messaging in the express checkout gateway plugin. Rather we should focus our efforts on making the onboarding to PayPal Payments as easy as possible through that extension. 

Therefore, we can delete this message once PayPal Payments is installed.

achyuthajoy:
Thanks @LevinMedia. I'll remove the message once the plugin is installed & configured.

LevinMedia:
@achyuthajoy Just to be clear - I think we can remove the message as soon as the new gateway is _installed_, before it's configured. We should rely on the new gateway to drive the configuration process :)

achyuthajoy:
Sure @LevinMedia. I'll make the change. 😄

includes/class-wc-gateway-ppec-plugin.php
achyuthajoy: Removed

assets/js/admin/ppec-upgrade-notice.js
achyuthajoy: Changes updated

woocommerce-gateway-paypal-express-checkout.php
achyuthajoy: Changes updated

templates/paypal-payments-upgrade-notice.php
achyuthajoy: Changes updated

templates/paypal-payments-upgrade-notice.php
achyuthajoy: Changes updated

achyuthajoy:
@jorgeatorres The requested changes are updated. Couple of new changes based on https://github.com/woocommerce/woocommerce-gateway-paypal-express-checkout/pull/866#issuecomment-860837916

1. Notice would be hidden if PayPal Payments is installed and active
2. `Configure PayPal Payments now` button removed

I've also updated the PR testing instructions to remove `Configure PayPal Payments now`.

assets/js/admin/ppec-upgrade-notice.js
```diff
@@ -0,0 +1,43 @@
+;(function ( $, window, document ) {
+	'use strict';
+
+	// Check whether PayPal Payments is installed.
+	let is_paypal_payments_installed = false,
+		is_paypal_payments_active = false;
+
+	const targetElement = $( 'tr[data-slug="woocommerce-paypal-payments"]' );
```
jorgeatorres, line 8: Even though it is unlikely, it could happen that a user has renamed the PayPal Payments directory to something else (say `woocommerce-paypal-payments2/`) for some reason.
Given the PHP uses `woocommerce-paypal-payments/woocommerce-paypal-payments.php` when generating the activation button we should probably look for a `data-plugin` attribute with that same value instead of relying on `data-slug`. That way, we can leave possible user plugin modifications undisturbed by installing/activating only the official (vanilla) extension.

To test:

1. Install PayPal Payments but keep it disabled.
2. Rename its directory.
3. Click "Activate PayPal Payments now" inside the PPEC message.
4. See a "Plugin file does not exist" error.

What are your thoughts?

assets/js/admin/ppec-upgrade-notice.js
```diff
@@ -0,0 +1,43 @@
...
+	if ( targetElement.length ) {
+		is_paypal_payments_installed = true;
+
+		if ( targetElement.hasClass( 'active' ) ) {
+			is_paypal_payments_active = true;
+		}
+	}
+
+	// Hide notice/buttons conditionally.
+	if ( is_paypal_payments_installed && is_paypal_payments_active ) {
```
jorgeatorres, line 22: Instead of trying to hide things when the JS is loaded, could we maybe hide the notice and buttons _by default_ using the CSS file (or even a `style: display: none;` inline attribute if necessary) and only show the relevant buttons if needed? Otherwise, even though it's not easy to see, the screen would flicker while the initial buttons are shown/hidden.

This is not critical. Feel free to skip this request completely if you're busy with other stuff.

jorgeatorres (changes requested):
Hi @achyuthajoy!

Thanks for making the changes 💯. This is looking good and working like a charm.
I left some minor feedback. Once it is addressed, we'll be good to merge.

Let me know what you think!

assets/js/admin/ppec-upgrade-notice.js
achyuthajoy: Good catch @jorgeatorres. The easiest fix I can come up with is to update the activation URL on notice via JS by copying the original plugin activation URL.

assets/js/admin/ppec-upgrade-notice.js
achyuthajoy: Made the entire buttons area hidden by default. Gets displayed when JS initializes.

jorgeatorres (approved):
Good job on this one @achyuthajoy 💯 ! This works great and I really like how this final version of the code looks. I had my doubts about using JS to populate the activation link, but we can revisit later if necessary (though I doubt it as it works fine).

LGTM 🚀

